Issue #63, opened 2021-04-14
Add aspect ratio support to the resize directive


tonyfsullivan:
### Feature Request
I find myself wanting to resize images to multiple different sizes and formats for responsive and lazy loaded images, but always want to keep the same aspect ratio.

Ideally I could do something like this:

`./assets/awesome-image.jpg?width=300;600;900&height=169;338;506`

and imagetools would magically know I want that to be three 16:9 images.  Not surprisingly, I actually get back 9 images with all permutations of widths and heights.

### Proposed Transform
It'd be nice if we could specify either an aspect ratio and either width or height along the lines of

`./assets/awesome-image.jpg?width=300;600;900&aspect=1.77777777`

### Existing Workaround
There is a workaround here so this is more of a convenience request than anything else.  I can just split this out into three separate imports each specifying exactly one width and height.

JonasKruckenberg:
Hey @tonyfsullivan! Thanks for the idea, I definitely see the benefit. Two things I'm not sure about though:

1. Most people won't have the float representation of aspect ratios memorized and realizing that 1.7777 means 16:9 is a lot of mental gymnastics at a glance.

2. I'm unsure about the a new aspectRatio directive since it'd be mutually exclusive with width or height anyway, but I don't have a better idea right now anyway.

3. If we're going forward with the aspect directive, would the aspect ratio be calculated on based on the given size?
Example:
16:9 and a given width of 500 would mean resize the height to be 0.5625 times the width.
Now say we have been given the height instead and an aspect ratio of 16:9. Would this mean resize the width to be 0.5625 times the height? Most people would find this confusing since 16:9 is associated with landscape and 9:16 with portrait images.

Maybe you have some idea regarding this!

tonyfsullivan:
All excellent questions!  Here's my take on each question given my use case, but there may very well be a different API to cover more scenarios

1. Totally agree that float values like 1.77777 aren't ideal.  I hadn't thought through the exact implementation and got to a float just because I thought about doing something like `./image.png?w=1920&aspect=${16/9}` but that could actually run into issues with dynamic impmort variables.  The more common format of 16:9 might actually be a great API design for it as long as the colon character isn't a problem, `./image.png?w=1920&aspect=16:9`
2. Agree here as well, a separate aspectRatio directive separate from resize isn't ideal.  I was thinking aspect could be an extra optional parameter that the existing resize directive uses when both height and width aren't specified.
[...]

PR #71, opened 2021-04-24
#63: Add aspect ratio support to the resize directive

tonyfsullivan:
[Add aspect ratio support to the resize directive](https://github.com/JonasKruckenberg/imagetools/issues/63)

This change adds an `aspect` keyword used in the resize transform.

Use Cases:
- If aspect & width are provided, calculate height to match
- If aspect & height are provided, calculate width to match
- If only aspect is provided, calculate both height and width.  This will keep the the original image as large as possible, meaning only one dimension will be cropped to get the correct aspect ratio
- If height & width are both provided, ignore the `aspect` keyword

**Tests**
Test coverage added to the `core` package, covering the cases above as well as combinations with fit, background, position, and kernel

### DIFF
--- a/packages/core/src/transforms/resize.ts
+++ b/packages/core/src/transforms/resize.ts
@@ -4,7 +4,6 @@ import { getPosition } from './position'
 import { getKernel } from './kernel'
 import { getBackground } from './background'
 import { setMetadata, getMetadata } from "../lib/metadata";
-import { Sharp } from "sharp";
 
 export interface ResizeOptions {
     width: string
@@ -81,6 +80,7 @@ export const resize: TransformFactory<ResizeOptions> = (config, ctx) => {
 
         setMetadata(image, 'height', finalHeight)
         setMetadata(image, 'width', finalWidth)
+        setMetadata(image, 'aspect', aspect || metaAspect)
 
         return image.resize({
             width: finalWidth,
--- a/packages/core/src/transforms/resize.ts
+++ b/packages/core/src/transforms/resize.ts
@@ -4,12 +4,33 @@ import { getPosition } from './position'
 import { getKernel } from './kernel'
 import { getBackground } from './background'
 import { setMetadata, getMetadata } from "../lib/metadata";
+import { Sharp } from "sharp";
 
 export interface ResizeOptions {
     width: string
     w: string
     height: string
     h: string
+    aspect: string
+}
+
+function parseAspect(aspect?: string) {
+    if (!aspect || !aspect.split) {
+        return undefined;
+    }
+
+    const [width, height] = aspect.split(':')
+
+    if (!width || !height) {
+        return undefined
+    }
+
+    const widthInt = parseInt(width)
+    const heightInt = parseInt(height)
+
+    return widthInt && heightInt && widthInt > 0 && heightInt > 0
+        ? widthInt / heightInt
+        : undefined
 }
 
 export const resize: TransformFactory<ResizeOptions> = (config, ctx) => {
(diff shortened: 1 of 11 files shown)


packages/core/src/transforms/resize.ts
```diff
@@ -4,12 +4,33 @@ import { getPosition } from './position'
 import { getKernel } from './kernel'
 import { getBackground } from './background'
 import { setMetadata, getMetadata } from "../lib/metadata";
+import { Sharp } from "sharp";
```
JonasKruckenberg, line 7: This import seems to be unnecessary 😄

packages/core/src/transforms/resize.ts
```diff
@@ -24,28 +45,46 @@ export const resize: TransformFactory<ResizeOptions> = (config, ctx) => {
...
+
+        const w = getMetadata(image, 'width')
+        const h = getMetadata(image, 'height')
 
-        if (!height) {
-            const w = getMetadata(image, 'width')
-            const h = getMetadata(image, 'height')
-            setMetadata(image, 'height', Math.round((width! / w) * h))
-            setMetadata(image, 'width', width)
+        const metaAspect = w / h
```
JonasKruckenberg, line 61: I don't know why you'd need to use the metadata at this point, since -- and correct me if I'm wrong -- it should `getMetadat(image,'width')` should always return `undefined` at this point. Or am I missing something?

packages/core/src/transforms/resize.ts
```diff
@@ -24,28 +45,46 @@ export const resize: TransformFactory<ResizeOptions> = (config, ctx) => {
...
+            finalWidth = width
+        } else {
+            /* only height was provided, need to calculate width */
+            finalHeight = height
+            finalWidth = height * (aspect || metaAspect)
         }
 
+        setMetadata(image, 'height', finalHeight)
+        setMetadata(image, 'width', finalWidth)
+
```
JonasKruckenberg, line 84: Maybe you could add an aspect-ratio metadata property just for good measure, I could image that being useful to people importing the image, sure I'd be a floating point number then, but people could make use of that still.
```suggestion
setMetadata(image,'aspect',aspect)
```

JonasKruckenberg (changes requested):
First off, great work man! Thanks a lot for the help, I added a few minor remarks regarding the resize.ts file but expect for that, outstanding work 👍

packages/core/src/transforms/resize.ts
tonyfsullivan: Hah, good catch!  I started with a helper that took in the Sharp image, removed the helper but missed this type import

packages/core/src/transforms/resize.ts
tonyfsullivan: I actually wasn't sure if this would work at first either, but the metadata is already populated from `sharp`

The case where only `aspect` is given takes advantage of it, looking at the metadata from `sharp` to decide which axis needs to be cropped down

packages/core/src/transforms/resize.ts
tonyfsullivan: Yep that's a great idea, done!

tonyfsullivan:
> First off, great work man! Thanks a lot for the help, I added a few minor remarks regarding the resize.ts file but expect for that, outstanding work +1

Thanks for the quick PR review!  Your repo has such a short list of open issues, I couldn't leave my feature request blowing up the issues count by 25% :laughing: 

Think I have all your comments covered in the latest changes, and merged in the latest changes from `main`

packages/core/src/transforms/resize.ts
JonasKruckenberg: Oh duh 🤦‍♂️ yes you're absolutely right

JonasKruckenberg (approved):

PR #74, opened 2021-04-26
#63: Adding `ar` shorthand for aspect and support for numeric values for aspect

tonyfsullivan:
Integrating feedback from [polarathene's comment](https://github.com/JonasKruckenberg/imagetools/issues/63#issuecomment-826483059), better matching Cloudinary's support for [aspect ratio](https://cloudinary.com/documentation/transformation_reference#ar_aspect_ratio)

This adds `ar` as a shorthand for `aspect` in the resize transform, e.g. `ar=16:9` is the same as `aspect=16:9`

This also adds support for numeric aspect values representing the width divided by height, e.g. `aspect=1.5` is the same as `aspect=3:2`


packages/core/src/transforms/resize.ts
```diff
@@ -46,7 +54,9 @@ export const resize: TransformFactory<ResizeOptions> = (config, ctx) => {
             : undefined
     const aspect = width && height
         ? width / height
-        : parseAspect(config.aspect)
+        : config.aspect
```
JonasKruckenberg, line 57:
```suggestion
        : parseAspect(config.aspect || config.ar)
```
why not use this?

JonasKruckenberg (changes requested):
You should also add the `ar` property to the `ResizeOptions` interface, but the rest looks good 👍

packages/core/src/transforms/resize.ts
tonyfsullivan: Great catch on my missing `ResizeOptions` for `ar`

> why not use this?
Yeah, there's no good reason... I skipped right past that cleaner `||` option and went straight for a nested ternary.  Just pushed a change to clean that up

JonasKruckenberg (approved):

packages/core/package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "dev": "rollup -cw",
     "build": "rollup -c",
-    "test": "jest",
+    "test": "jest src/transforms/__tests__/resize.spec.ts",
```
JonasKruckenberg, line 30:
```suggestion
    "test": "jest",
```
You probably forgot to remove this before pushing 😉

JonasKruckenberg (changes requested):

packages/core/package.json
tonyfsullivan: Shoot, sorry about the sloppy check-ins here!  Thanks for catching this

JonasKruckenberg (approved):

PR #90, opened 2021-05-07
Readme fix

cxa:


JonasKruckenberg (approved):

JonasKruckenberg:
I first thought having a license for each module would be find, but I guess you're right, having a license for the whole project is important too.

Issue #210, opened 2021-10-13
can't run tests locally


benmccann:
I ran this:
```
sudo apt install git-lfs
yarn install
yarn build
yarn test
```

I got:
> Test Suites: 14 failed, 11 passed, 25 total
Tests:       71 failed, 173 passed, 244 total

They were all failing with the message `Input file contains unsupported image format`

Is there some step I might be missing?

Issue #139, opened 2021-06-11
Support multiple values of the same key in defaultDirectives


qelix:
Hey @JonasKruckenberg,

thanks for the great project! Probably like most people, I want to use this plugin to create a bunch of differently sized images that are later served as an image with a `srcset` tag. Because I don't want to specify all of the breakpoints individually for each image, I found out that there's the `defaultDirectives` setting in the Vite plugin options. So the idea is to store a variable containing all of the breakpoints, which is then picked up by the Vite config and automatically applied to every imported image. 

However, the setting doesn't seem to support more than one value for each property at a time -- I tried passing the string exactly like you would in an import statement (`width=3840;2560;1920` etc.), but the plugin only seems to accept the first number it gets. The exact config I used looks like this:

```js
plugins: [
  imagetools({
    defaultDirectives: { width: imageWidths.join(";"), format: 'webp' },
  }),
],
```

Is there any way to solve this? I also thought about using dynamic import statements, but I never really got them to work properly. I really like your project, but the design decision to make everything a directive you have to spell out for each individual import makes it quite hard to outsource a common configuration. 

Your help would be appreciated :)

JonasKruckenberg:
Sooo, I thought this was gonna come up at some point ;) 
The way imagetools works is it first parses the url and generates an object from that. It then generates what I call targets, one for each image that is gonna be produced.
The issue is that in your case the defaultDirectives get applied **after** the last step above. That's why it's not working.
I am however very unhappy with the way the API works right now so I'm doing (yet another *sigh*) rewrite and I'll make sure to address this issue too!

JonasKruckenberg:
Alright let's think about this again, back in June I promised an API redesign that sadly went nowhere (for different reason but mostly time 😕) so lets move this forward and think of a way we can approach this without a redesign.
My first idea was to change `defaultDorectives` from
```ts
type defaultDirectives = Record<string, string> | ((id: string) => Record<string, string>)
```
to this
```ts
type defaultDirectives =  URLSearchParams | ((id: string) =>  URLSearchParams)
```
and merge them with the  `URLSearchParams` from the actual Url. This would resolve #216 since configs would be applied before resolving and allow for multiple values for the same key.
wdyt @Porges @qelix @jugglingcats would that solve you use case?

Porges:
[...]

PR #215, opened 2021-10-29
feat: Add ability to override resolveConfigs in vite plugin

jugglingcats:
- **Quick Checklist**

This pull request relates to #139. It adds ability to supply a custom `resolveConfigs`.

If you don't set property in config everything works the same as before. You can supply a function that will be called in place of the default `resolveConfigs`. If this function doesn't return a value, the existing `resolveConfigs` is called, otherwise the result of the custom `resolveConfigs` is used.

Below is an example of a custom implementation. If you add `?glowsite=<preset>` to an image import it will generate images with preset widths in webp format.

```
const defaultPresets = {
    default: {widths: [300, 800, 1200]}
}

export function glowsiteImageToolsPresets(pluginConfig) {
    const presets = pluginConfig?.presets || defaultPresets

    return function (config) {
        const glowsite_config = config.find(([key]) => key === "glowsite")
        if (glowsite_config) {
            const [, [value]] = glowsite_config
            const preset_name = value.trim().length ? value.trim() : "default"
            const preset = presets[preset_name] || defaultPresets.default
            const widths = preset.widths

            return widths.map(width => ({
                width,
                webp: ""
            }))
        }
    }
}
```

```
// vite.config.js
...
export default {
    plugins: [
        react(),
        imagetools({
            resolveConfigs: glowsiteImageToolsPresets({default: {widths: [100,200,300]}})
        }),
...
```

  PR?)

No (intentional) breaking changes.

- **Other information**:

If you think this PR might be accepted I can add tests, docs, changeset, etc.

I had trouble running the tests on Windows (and on Ubuntu actually) after clean checkout of your repo, so might need some help resolving that before I can add tests.

^--- issue on Ubuntu is probably git-lfs issue (similar to #210) in which case I can workaround Windows issue

### DIFF
--- a/.changeset/resolve-configs.md
+++ b/.changeset/resolve-configs.md
@@ -1,7 +1,6 @@
 ---
-'imagetools-core': patch
-'rollup-plugin-imagetools': patch
-'vite-imagetools': patch
+'rollup-plugin-imagetools': minor
+'vite-imagetools': minor
 ---
 
 feat: allow override of resolveConfigs
--- a/.changeset/resolve-configs.md
+++ b/.changeset/resolve-configs.md
@@ -1,0 +1,7 @@
+---
+'imagetools-core': patch
+'rollup-plugin-imagetools': patch
+'vite-imagetools': patch
+---
+
+feat: allow override of resolveConfigs
--- a/docs/guide/extending.md
+++ b/docs/guide/extending.md
@@ -118,6 +118,9 @@ Sometimes it's useful to be able to override the resolution of image configs at 
 to implement presets using shortnames that generate multiple images. This resolution is done before the config
 is passed to the directives. To implement a custom scheme, pass a function as `resolveConfigs` plugin options.
 
+The function should return a new config array. If a plugin needs to modify the config that would have been used,
+it can call the default `resolveConfigs` function which is part of the public API.
+
 Below is an example custom `resolveConfigs` to implement site preset widths which are always webp format.
 
 ```ts
--- a/docs/guide/extending.md
+++ b/docs/guide/extending.md
@@ -111,3 +111,63 @@ function customDirective(): ImageTransform {
 ## Custom Output Formats
 
 _TODO_
+
+## Custom config resolution
+
+Sometimes it's useful to be able to override the resolution of image configs at a low level, for example
+to implement presets using shortnames that generate multiple images. This resolution is done before the config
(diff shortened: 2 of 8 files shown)


JonasKruckenberg:
Looks good to me! This is in many regards better than all the solutions I have thought of, since it actually works 😄.
I'd propose changing it to a function that receives the configs array and must return a config array 🤔 so a preset could also modify and not replace the configs.
But besides that I'll be happy to include this! Please make sure to add tests and all that 👍🏻 
Thanks for your patience

JonasKruckenberg:
Actually, I just realized that this might be a bit redundant if my proposal in #139 gets merged. Or is there something I am missing?

jugglingcats:
Hi @JonasKruckenberg, have ported to rollup plugin, added a test in both and added a section to the guide.

I made a comment in #139 which I hope explains why this option provides more flexibility over the `defaultConfigs` route.

I did not implement your request to accept and return a config array. Might just be me being lazy but I feel if you are overriding `resolveConfigs` you probably want complete control and don't care what the default resolution would be. And if you really need it you can call the core `resolveConfigs` easily enough. If you feel strongly I can look at it again!

JonasKruckenberg:
Good point, the core `resolveConfigs` function is part of the public API. Could you just add a short sentence in the docs to mention this?
And the changeset it just a markdown file in the `.changeset` folder, but you can also run `yarn changeset add` in the project directory and the cli will guide you through the process. Make to sure to include minor bumps for the vite and rollup plugin though!

JonasKruckenberg:
Sorry, forgot to update the PR check workflow to use node 16.x instead 15.x. Can you quickly rebase against main again so we can run the checks? Thanks!

JonasKruckenberg:
Awesome work! Sorry for adding this so late in the process, but what do you think of renaming it to `overrideConfigs`? I guess that'd make it clearer what that hook is useful for 🤔

jugglingcats:
Hi I think that might be confusing for a couple of reasons, a) people looking at the code wouldn't immediately make a link between `overrideConfigs` and `resolveConfigs`, b) `overrideConfigs` doesn't immediately shout function to me, I would expect it to be an override value.

I would vote for leaving as it is, or consider changing to `overrideResolveConfigs`...! Happy to make that change if you agree.

JonasKruckenberg:
yeah, that makes sense to me!

JonasKruckenberg (approved):

PR #218, opened 2021-11-05
fix: clone image in vite plugin to avoid dev server issue (#140)

### LINKED ISSUE #140: Generating multiple sizes only uses the last size for the entire set
Hey, another one from me ;)

I discovered a serious bug that actually completely breaks the source set generation. If you generate multiple images by chaining multiple widths together (e.g. `import srcsetAvif from '../example.jpg?w=500;700;900;1200&avif&srcset'`, as per your example), it actually only applies the last width on every image in the whole set!

To reproduce: 

1. Clone this repo and `cd` into `examples/vite-simple`
2. Edit `main.js` and add ridiculously low widths as the last params: 

```js
// ...
import srcsetAvif from '../example.jpg?w=500;700;900;1200;20&avif&srcset'
import srcsetWebp from '../example.jpg?w=500;700;900;1200;20&webp&srcset'
// ...
```
3. `yarn && yarn dev`
4. Open the page in your browser, open the devtools and reload
5. In your network tab, you'll now see the exact file that was loaded - for me it had an ID of `e78e253f7e1877238273f8472026215d31af4221` (you guessed it: it's only 20px wide 😉)
6. When inspecting srcset of the the image, however, you'll notice that the loaded ID was supposed to be 1200px wide (depending on your viewport, of couse): `/@imagetools/e78e253f7e1877238273f8472026215d31af4221 1200w,`. The one with 20px has an entirely different ID.

So, even though the 20px wide image was not selected by the browser, it still only received a 20px wide one. It looks like it's always the last argument that decides the size of all the images in the set. All of the metadata is still generated according to the parameters. Each URL is just pointing to an image the size of the last parameter.

jugglingcats:
- **Quick Checklist**

Fixes #140 

None

  PR?)

No

- **Other information**:

Am not sure if cloning an image has performance implications. Perhaps could be optimised by only cloning if there is more than one image format in the array, but didn't see worth premature optimisation.

### DIFF
--- a/.changeset/fix140.md
+++ b/.changeset/fix140.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+fix: clone sharp image in vite plugin to avoid reuse issue
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -65,7 +65,7 @@ export function imagetools(userOptions: Partial<PluginOptions> = {}): Plugin {
             : pluginOptions.defaultDirectives
 
         const { transforms } = generateTransforms({ ...defaultConfig, ...config }, transformFactories)
-        const { image, metadata } = await applyTransforms(transforms, img, pluginOptions.removeMetadata)
+        const { image, metadata } = await applyTransforms(transforms, img.clone(), pluginOptions.removeMetadata)
 
         generatedImages.set(id, image)
 


jugglingcats:
I also could not think of a way to unit test this...

JonasKruckenberg:
Same as with #215 can you quickly rebase this against main?

JonasKruckenberg:
Also please run `yarn changeset add` and add a `fix`bump for the vite plugin

JonasKruckenberg (approved):

JonasKruckenberg:
Thanks a lot! 🙏

PR #220, opened 2021-11-09
Fix path to typings

mape:
Fixes the error "Cannot find module 'vite-imagetools' or its corresponding type declarations."


JonasKruckenberg:
Hey, thanks for the PR! Could you please add this fix to the rollup too?

JonasKruckenberg:
awesome thanks for the fix 👍🏻

Issue #260, opened 2021-12-02
Easier creation of `<picture>` tags


benmccann:
I just tried the [experimental `srcset` option](https://github.com/JonasKruckenberg/imagetools/blob/main/docs/spec.md#srcset)

For my example, it created one large `srcset`:
```
/@imagetools/af1896684ba60c3fa5b12a0609e28b3fce179657 1440w, /@imagetools/d9f7937f7986fa94eac8fbdcd50acb2ec8db14a2 1440w, /@imagetools/0f418c158ad24a2c7a101f8847a727731a0c1b55 1440w, /@imagetools/c6ee3dba26e4ad983a4ae2ea2a69076492153ffc 800w, /@imagetools/84620b40939e5c4ca0a61b4724c29a8ee2057f95 800w, /@imagetools/91bce48b94bc945a6df13919b8ec4b7ed8521c17 800w
```

But really what I need is something much more complex:
```
<picture class="s-Ou-fLRYjMacr">
   <source type="image/avif" srcset="/@imagetools/425c37d1844269144a629fd6607654a047acb328 2x, /@imagetools/0ac134d9b8e160cc8352e48344071d15855e0868 1x" class="s-Ou-fLRYjMacr">
   <source type="image/webp" srcset="/@imagetools/9b27e371177bb36997e6b36e9d0ebee369889fa6 2x, /@imagetools/ddc6d76b452971e6f9b5ac6e5f921b04c2b8b54d 1x" class="s-Ou-fLRYjMacr">
   <source type="image/png" srcset="/@imagetools/2e89026d6e552fae215d9952d95d67ff974960fa 2x, /@imagetools/ef93de319f4f4f4ad4f7b555ab260d4ffd31d7c4 1x" class="s-Ou-fLRYjMacr">
   <img src="/@imagetools/2e89026d6e552fae215d9952d95d67ff974960fa" alt="SvelteKit illustration" class="s-Ou-fLRYjMacr">
</picture>
```

A couple things would be helpful here:
- Output grouped by format
- Normalizing the output format for use in the `type` attribute (always return `jpeg` instead of sometimes `jpg` and sometimes `jpeg`)

We could fallback to the largest image of the user-provided type.

[...]

PR #369, opened 2022-08-25
feat: add picture and source output formats

benmccann:
- **Quick Checklist**

* [X] I have written new tests, as applicable (for bug fixes / features) - will do that tomorrow
* [X] I have added a changeset, if applicable

Feature

Closes https://github.com/JonasKruckenberg/imagetools/issues/260

No

- **Other information**:

### DIFF
--- a/.changeset/polite-moles-live.md
+++ b/.changeset/polite-moles-live.md
@@ -1,0 +1,5 @@
+---
+'imagetools-core': patch
+---
+
+feat: add picture and source output formats
--- a/docs/modules/core_src.md
+++ b/docs/modules/core_src.md
@@ -76,10 +76,12 @@
 - [metadataFormat](core_src.md#metadataformat)
 - [normalize](core_src.md#normalize)
 - [parseURL](core_src.md#parseurl)
+- [pictureFormat](core_src.md#pictureformat)
 - [resize](core_src.md#resize)
 - [resolveConfigs](core_src.md#resolveconfigs)
 - [rotate](core_src.md#rotate)
 - [setMetadata](core_src.md#setmetadata)
+- [sourceFormat](core_src.md#sourceformat)
 - [srcsetFormat](core_src.md#srcsetformat)
 - [tint](core_src.md#tint)
 - [urlFormat](core_src.md#urlformat)
@@ -269,6 +271,8 @@ ___
 | :------ | :------ |
 | `meta` | [`OutputFormat`](core_src.md#outputformat) |
 | `metadata` | [`OutputFormat`](core_src.md#outputformat) |
+| `picture` | [`OutputFormat`](core_src.md#outputformat) |
+| `source` | [`OutputFormat`](core_src.md#outputformat) |
 | `srcset` | [`OutputFormat`](core_src.md#outputformat) |
 | `url` | [`OutputFormat`](core_src.md#outputformat) |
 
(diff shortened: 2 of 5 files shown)


JonasKruckenberg (changes requested):
Okay just a couple of small things:

1. add a test case for the formats here https://github.com/JonasKruckenberg/imagetools/blob/main/packages/core/src/__tests__/output-formats.test.ts
2. add docs to this file https://github.com/JonasKruckenberg/imagetools/blob/main/docs/directives.md

But I actually have one bigger question: What benefit does the `source` output format has over using `metadata=src;width`? I'm not against it, just wondering. The fact that the `metadata`  format accepts a whitelist is a little advertised feature too.

benmccann:
I've added docs and tests now

>What benefit does the source output format has over using metadata=src;width?

I added some docs to the type explaining: https://github.com/JonasKruckenberg/imagetools/pull/369/files#diff-4c2458380e50c967c70d7afe819dc7412180f9f817043f27441de3498e4d0479R31

zerodevx:
Hi all, since we're on the topic of `outputFormats`, I was wondering if we can `async` this too:

https://github.com/JonasKruckenberg/imagetools/blob/b84387434d60e7ea53c6fb806134e8f1d24e6243/packages/vite/src/index.ts#L109-L113

Like so:

```js
return dataToEsm(await outputFormat(outputMetadatas), { ... }
```

It's possible that an `outputFormat` is an `async` op, for eg. generating the `base64` of a low-quality image placeholder. This enables a consumer to extend in `vite.config.js`:

```js
const config = {
  plugins: [sveltekit(), imagetools({
    extendOutputFormats: (defaults) => ({
      ...defaults,
      inline: () => async (metas) => {
        const out = []
        for (const { image, format } of metas) {
          const buffer = await image.toBuffer()
          out.push(`data:image/${format};base64,${buffer.toString('base64')}`)
        }
        return out.length === 1 ? out[0] : out
      }
    })
  )]
}
```

Does this make sense?? 🤔

benmccann:
@zerodevx it sounds like a good idea, but I'd rather do it separately than expand the scope of this PR. can you file a new issue for it?

JonasKruckenberg (approved):

s3812497:
Hi @benmccann . Is there a way for the `picture` directive to output the height as well?

I'm trying to migrate to this from the [old solution provided here](https://github.com/JonasKruckenberg/imagetools/issues/260#:~:text=commented-,on%20Jan%2011,-i%20did%20something).

Currently, when using `picture`, it's difficult to set the height to calculate the aspect ratio of the image. It also overrides the `metadata` directive, so it can't be mixed.

I've been using this setup with SvelteKit and it's almost perfect:
```js
// vite.config.js
const config = {
	plugins: [
		sveltekit(),
		imagetools({
			defaultDirectives: (id) => {
				if (id.searchParams.has('optimize')) {
					return new URLSearchParams(
						'w=1920;1366;780;414&format=avif;webp;jpg&meta=width;height;src;format'
					);
				}
				if (id.searchParams.has('placeholder')) {
					return new URLSearchParams('w=54&blur=2&jpg');
				}
				return new URLSearchParams();
			},
		})
	]
};

// +page.svelte
import Image1 from '$images/dog.jpg?optimize';
```
with an `Image.svelte` component[ adapted from this example.](https://github.com/JonasKruckenberg/imagetools/issues/260#:~:text=commented-,on%20Jan%2011,-i%20did%20something)

benmccann:
@s3812497 can you share the `Image` component you're using to show how you are using the `height`? It's probably best to file the request as a new issue so that it doesn't get lost in a closed PR

PR #399, opened 2022-09-22
fix: let vite create final paths

pzerelles:
- **Quick Checklist**

* [n/a] I have written new tests, as applicable (for bug fixes / features)
* [n/a] Docs have been added / updated (for bug fixes / features)

This PR fixes the paths of images in the Vite build result. Before the filename was added to `viteConfig.base` which leads to relative urls that don't work anymore in nested paths.

Instead of inserting the final path in imagetools, the existing placeholder logic from Vite is used and Vite will do the replacement.

  PR?)

- **Other information**:

### DIFF
--- a/.changeset/two-games-kiss.md
+++ b/.changeset/two-games-kiss.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+Let vite create correct paths in build result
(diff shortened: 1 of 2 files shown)


pzerelles:
Does not work with srcset yet...

PR #430, opened 2022-10-28
Upgrade to Rollup 3

benmccann:
I didn't include a changeset. The compiled output is probably slightly different, but it shouldn't be user-visible


JonasKruckenberg:
Nice!

Issue #317, opened 2022-01-31
the ability to convert Gif to Webp animated?


EnderPicture:
Hi,

I see that it is possible for Sharp to convert Gif to animated Webp files in their [doc](https://sharp.pixelplumbing.com/api-constructor). Is it possible in image tools? currently, the conversion works but only to a static image.

Thanks!

JonasKruckenberg:
Your link seems to just go to their homepage and I couldn't find the specific option you mentioned.

Maybe you can provide a quick reproduction with a gif that would be converted into an animated webp normally but doesn't get converted when using imagetools so I can can check it out!

millette:
I think he's referring to `options.pages`.

EnderPicture:
Sorry for the delay, I was referring to this section
```js
// Convert an animated GIF to an animated WebP
await sharp('in.gif', { animated: true }).toFile('out.webp');
```

TomBor:
From the [constructor page ](https://sharp.pixelplumbing.com/api-constructor) of Sharp website, there are 2 references to animated file in the options
+ `options.pages` [number ](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number) number of pages to extract for multi-page input (GIF, WebP, AVIF, TIFF, PDF), use `-1` for all pages. (optional, default `1`)
+ `options.animated` [boolean ](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean) Set to `true` to read all frames/pages of an animated image (equivalent of setting `pages` to `-1`). (optional, default `false`)

Would be great to have an option for webp animated or not in imagetools.

Amerlander:
Any plans to implements this? 

I guess line `6` in `util.ts` has to be extended to receive the option to enable animations:  
```js
return sharp(path, { animated: true })
``` 

https://github.com/JonasKruckenberg/imagetools/blob/c70b97e22256550aa9eac4bb7c7e9063c881ac00/packages/core/src/util.ts#L6

PR #454, opened 2022-12-09
enable animations per flag

Amerlander:
~This PR will set `animated: true` for all gif files that are processed.~
~Converting an animated gif to apng, webp or aviv will result in an animated image. Converting to jpeg wont.~

Related issue: https://github.com/JonasKruckenberg/imagetools/issues/317

~Might be better to have this as an option (for each image or in global config  for all files / per file type). But was not sure whats the best way to pass an option to `loadImage()`.~

**Update:** This PR introduces an `animated` flag, which enables imagetools to read all frames of an animated image and return an animated image for suported formats.

I have tests included, but since the test environment does not seem to work on my pc I could not verify them yet.


chojnicki:
@JonasKruckenberg is this repo still maintained? :(

neil3000:
This really should be merged

JonasKruckenberg:
Okay so @Amerlander if you could add some tests that would be much appreciated. Basically just ensuring that images that should be animated stay animated. Then we can merge this PR at long last.

To answer your question @chojnicki (even though this really is not the right format for this) maintenance on this crate is slow and new features are hard to get in here. This is down to time: I have many other projects, many other packages I maintain and most of these are more interesting/rewarding/important than this plugin. I just don't have enough time (don't we all?).
So, if you're up for the challenge and you wanna help bring optimized images to more websites feel free to help with maintenance. I would be overjoyed and help wherever possible.

benmccann:
The reason I've been holding off on this PR is that it's not clear to me what the impact of making a non-animated gif into an apng, etc. would be. If we're making files larger by changing them from a non-animated to an animated format then this may not be the best solution. It would be helpful if someone could show the file size change of converting a png to apng.

neil3000:
> The reason I've been holding off on this PR is that it's not clear to me what the impact of making a non-animated gif into an apng, etc. would be. If we're making files larger by changing them from a non-animated to an animated format then this may not be the best solution. It would be helpful if someone could show the file size change of converting a png to apng.

Maybe make it as an optional parameter? Something like `?anim`, that way developers can choose themselves wether it animate or not

Amerlander:
@benmccann I did some basic test:

- **Good**: Converting a format without animation into any other format is always the same size with and without `--animated`.
- **Good**: Converting to `png` will never get you an `apng`. You have to set target explicit to `apng` to get one.
- **Good**: Converting animated image into an animatable format will result in an static image without `--animated`
- **Bad**: Converting an animated image using `--animated` into a format that can not hold animations might result in a static image that holds all frames (a 3-frames-gif might result in a png that has tripple the height showing all frames on top of each other).

The current aproach of this PR to filter for `gif` files does not makes much sense then, since it ism ore about the target format. I think there are several better solutions:

- **Enable `--animated` trough an optional parameter** like @neil3000 suggest. Current behaviour is not changed and its up to the user to enable animations if desired.
- **Enable `--animated` when the target formats support animations:** Should work well in most cases, static formats will only show the first frame, animated formats will show the animation.
- **Always enable `--animated`:** Not smart, but as `imagetools` might mostly be used to optimize images for the web the target formats might mostly be `webp` and `aviv` which supports animations. No changes for static to static transformations, but animated to static will get unwanted results.

Always enabling `--animated` is easy, but I was not able to see how I could get informations about parameters like output format in `loadImage()`. https://github.com/JonasKruckenberg/imagetools/blob/83bbbc06753d405a6c9dda98c1347d23e8a7033e/packages/core/src/util.ts#L6

JonasKruckenberg:
Okay, I think option 3 (always enable) is squarely off the table given what you described and following the spirit of image tools other transforms having a `animated` flag would be best I think. Having it enabled for output formats that support animation would be neat too, but it raises the question of disabling that behaviour and since all other flags are positive (meaning they enable features) adding a negative flag (like `no-anim`) would feel weird.

Amerlander:
I changed this PR now to introduce an `animated` flag

```ts
/**
* Set to true to read all frames/pages of an animated image (GIF, WebP, TIFF), equivalent of setting pages to -1.
* @default true
*/
animated: boolean
  ```

I also added tests, but since the test are not executeable even without my changes on my pc I couldnt run them.

(Maybe I do it the wrong way?
Executing `pnpm run test` in `/imagetools/` fails with ` ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  imagetools-core@3.3.1 test: 'vitest run' Exit status 3221225477`
In `/imagetools/packages/core/` claims `Test failed`, but all tests are checked green 
And executing in `/imagetools/packages/vite/` says ` Failed to resolve entry for package "imagetools-core". The package may have incorrect main/module/exports specified in its package.json.`
)

**Comment by benmccann** [comment]
@Amerlander that's the error message you get if you run the tests without building first. If you run `pnpm build` and then `pnpm test` it should work. You'll also need to run `pnpm format` as lint is currently failing

File: packages/vite/src/types.ts [hunk 1]
```diff
@@ -77,4 +77,10 @@ export interface VitePluginOptions {
    * @deprecated
    */
   force?: boolean
+
+  /**
+   * Set to true to read all frames/pages of an animated image (GIF, WebP, TIFF), equivalent of setting pages to -1.
+   * @default false
+   */
+  animated: boolean
```
benmccann, line 85: the option @neil3000 suggested was a per-image option (aka a directive) passed as a query string: https://github.com/JonasKruckenberg/imagetools/blob/main/docs/directives.md.  

This is a per-project option, which is probably not what we want. We can use `defaultDirectives` to turn it entirely on or off for a project: https://github.com/JonasKruckenberg/imagetools/tree/main/packages/vite#defaultdirectives

neil3000: Indeed. I think it's better for this parameter to be a per-image option, as most of the time the developer know if the image he included on the site is animated or not.

If it's dynamic content (like user generated content) they can choose to implement their own logic :)

Amerlander: thank you for the feedback. I will look into this the next days. I found there are also some things I got wrong for the tests, so it might be better I close this PR for now and will come up with a new one when I worked myself into the logic of this repo and managed to implement this feature and the tests.

PR #461, opened 2022-12-23
feat: add withoutEnlargment + withoutReduction

### LINKED ISSUE #460: Add `withoutEnlargement ` and `withoutReduction`
Add support for Sharp's `withoutEnlargement` and `withoutReduction` parameters. This could be a part of the ["Improve resize directive"](https://github.com/JonasKruckenberg/imagetools/projects/1#card-60660311) card on the roadmap

A cursory look at the source code and it doesn't look like too much trouble to add. I'm happy to take a stab at this in a PR.

sawyerclick:
- **Quick Checklist**

Added boolean `withoutEnlargement` and `withoutReduction` flags to prevent image resizing if active and the calculated with or height exceeds or subceeds the original size.

Default behavior remains the same but users can add `&withoutEnglargment` or `&withoutReduction` to prevent unwanted resizing. Images are still created as Sharp sees fit.

  PR?)
No breaking change as `withoutEnlargement` and `withoutReduction` are opt-in per Sharp's API.

- **Other information**:
Closes #460 and touches #257. 

1. Check that either flag is present (`'' || 'true'`)
2. Check that the finalWidth and finalHeight do not subceed/exceed according to each parameter
3. If either is flagged, revert to original image size

Sharp's [fastShrinkOnLoad](https://sharp.pixelplumbing.com/api-resize) could be added while we are here.

### DIFF
--- a/packages/core/src/transforms/resize.ts
+++ b/packages/core/src/transforms/resize.ts
@@ -92,7 +92,7 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       finalAspect = originalAspect
 
       console.warn(
-        'withoutEnlargement or withoutReduction active. Image width, height and aspect ratio reverted to original values'
+        '[imagetools] withoutEnlargement or withoutReduction enabled. Image width, height and aspect ratio reverted to original values'
       )
     }
 
--- a/packages/core/src/transforms/resize.ts
+++ b/packages/core/src/transforms/resize.ts
@@ -90,6 +90,10 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       finalHeight = originalHeight
       finalWidth = originalWidth
       finalAspect = originalAspect
+
+      console.warn(
+        'withoutEnlargement or withoutReduction active. Image width, height and aspect ratio reverted to original values'
+      )
     }
 
     setMetadata(image, 'height', finalHeight)
--- a/packages/core/src/transforms/resize.ts
+++ b/packages/core/src/transforms/resize.ts
@@ -48,7 +48,12 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
   const withoutEnlargement = config.withoutEnlargement === '' || config.withoutEnlargement === 'true'
   const withoutReduction = config.withoutReduction === '' || config.withoutReduction === 'true'
 
-  if (!width && !height && !aspect) return
+  if (
+    (!width && !height && !aspect) ||
+    (config.withoutEnlargement && !withoutEnlargement) ||
+    (config.withoutReduction && !withoutReduction)
+  )
+    return
 
   return function resizeTransform(image) {
     // calculate finalWidth & finalHeight
@@ -82,7 +87,6 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       (withoutEnlargement && (finalHeight > originalHeight || finalWidth > originalWidth)) ||
       (withoutReduction && (finalHeight < originalHeight || finalWidth < originalWidth))
     ) {
-      // revert back to original sizes if either width or height exceeds or subceeds
       finalHeight = originalHeight
       finalWidth = originalWidth
       finalAspect = originalAspect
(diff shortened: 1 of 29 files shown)


packages/core/src/transforms/resize.ts
```diff
@@ -73,13 +77,24 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       finalWidth = height * (aspect || originalAspect)
     }
 
+    if (
+      (withoutEnlargement && (finalHeight > originalHeight || finalWidth > originalWidth)) ||
+      (withoutReduction && (finalHeight < originalHeight || finalWidth < originalWidth))
+    ) {
+      // revert back to original sizes if either width or height exceeds or subsceeds
```
benmccann, line 84: subsceeds seems like a typo

sawyerclick:
On second glance, it may be better to break withoutEnlargement and withoutReduction into their own files, similar to fit, position, kernel and background. No need to do the math on our end when Sharp will do it for us.

packages/core/src/transforms/resize.ts
```diff
@@ -43,8 +45,15 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
...
   const height = parseInt(config.height || config.h || '')
   const aspect = parseAspect(config.aspect || config.ar || '')
+  const withoutEnlargement = config.withoutEnlargement === '' || config.withoutEnlargement === 'true'
+  const withoutReduction = config.withoutReduction === '' || config.withoutReduction === 'true'
 
-  if (!width && !height && !aspect) return
+  if (
+    (!width && !height && !aspect) ||
+    (config.withoutEnlargement && !withoutEnlargement) ||
+    (config.withoutReduction && !withoutReduction)
```
JonasKruckenberg, line 54: I think you can omit the `config.withoutEnlargement` and `config.withoutReduction` here since you already took them into account above right?

packages/core/src/transforms/resize.ts
```diff
@@ -73,13 +83,26 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       finalWidth = height * (aspect || originalAspect)
     }
 
+    if (
+      (withoutEnlargement && (finalHeight > originalHeight || finalWidth > originalWidth)) ||
+      (withoutReduction && (finalHeight < originalHeight || finalWidth < originalWidth))
+    ) {
+      finalHeight = originalHeight
```
JonasKruckenberg, line 90: I might be worth adding a `console.warn` here to let the user know that the `withoutEnlargement` or `withoutReduction` case got hit, otherwise it might be surprising that the resize "didn't work"

JonasKruckenberg (changes requested):

JonasKruckenberg:
I also think your test cases only ever test the "positive case" so cases where e.g. `withoutEnlargement` wouldn't kick in anyway. Would be great if you could add some test cases that ensure the directive kicks in when the image is being over-enlarged and make sure that the image dimensions are equal to the originals.

sawyerclick:
> I also think your test cases only ever test the "positive case" so cases where e.g. `withoutEnlargement` wouldn't kick in anyway. Would be great if you could add some test cases that ensure the directive kicks in when the image is being over-enlarged and make sure that the image dimensions are equal to the originals.

Good point! I do have this line where I test multiple dimensions that do go under and over the original width. In the resulting images I do see for each that only the positive case is created, and the test for this is passing. The negative case image is not created since it is just the original image (at least here. in practice it would create the image for those with other non-resize directives such as `blur`, `invert`, etc.)

```ts
test('w/ multiple dimensions', async () => {
  // @ts-expect-error we know this is safe
  const { image } = await applyTransforms([resize({ withoutReduction: 'true', width: '300;900' }, dirCtx)], img)

  expect(await image.toBuffer()).toMatchImageSnapshot()
})
```

packages/core/src/transforms/resize.ts
sawyerclick: I couldn't find any other warnings in the app beyond the tests, so I've added one prefixed with `[vite-imagetools]`. Not sure if you have a standard you're looking for here

JonasKruckenberg:
I will change the warning message a bit, but I'll do that afterwards. Thanks for the PR!

JonasKruckenberg (approved):

benmccann:
@sawyerclick we're currently working on a new major version and I wanted to ask about your motivation for this PR. `withoutEnlargement` makes tons of sense to me and I'm thinking it may be beneficial to enable that by default. On the otherhand, I can't understand why anyone would use `withoutReduction`. Did you have a usecase for that or did you just add it for parity and to match the Sharp options?

My proposal would be to remove these two options, disable upscaling by default, and add a new option `upscale=true` that can be specified to enable upscaling. Would this be okay with your usecase?

CC @nick-vincent who filed an issue related to these options (https://github.com/JonasKruckenberg/imagetools/issues/510)

sawyerclick:
Howdy! `withoutReduction` was added solely for parity. I have not used it once myself but included it at the time as it mirrors `withoutEnlargement`. I do use `withoutEnlargement` often and would like it to persist. 

I'm all for simplifying this and only offering an `allowUpscale` parameter. Would this then be the only imagetools parameter that deviates from Sharp's method/param names?

PR #486, opened 2023-01-16
feat: add logging facilities

tlaundal:
~~Uses the `context.warn` function to issue errors from the transformers and prints these from the vite and rollup plugins. If the `silent` option is set, the warnings are not printed.~~

Logs messages through Rollup and Vite so message verbosity can be tuned.

I have tested this in a Vite project, but had no Rollup project to test in.

- **Quick Checklist**

Feature

Transforms now get access to a `logger` object through the context, which allows for logging at different levels (`info`, `warn`, `error`). 

Removes the existing `silent` functionality.

  PR?)
The way in which the warning is logged has changed. There might be a tiny chance that someone relies on parsing the plugin output to detect this, but this seems unlikely.

- **Other information**:
Output from vite before this PR:
Screenshot from 2023-01-16 18-17-38

Output from vite after this PR:
Screenshot from 2023-01-16 18-15-41


packages/core/src/transforms/__tests__/resize.spec.ts
```diff
@@ -9,12 +9,15 @@ import { resize } from '../resize'
 
 expect.extend({ toMatchImageSnapshot })
 
-describe('width', () => {
-  let dirCtx: TransformFactoryContext
-  beforeAll(() => {
-    dirCtx = { useParam: vi.fn, warn: vi.fn }
-  })
+let dirCtx: TransformFactoryContext
+beforeAll(() => {
```
tlaundal, line 13: I had to extract this to the top level, because I got interference between the different `describe` blocks. Not sure what was causing this, but perhaps parallel test running or something similar?

This was the simplest way I found of adressing the issue.

packages/core/src/transforms/__tests__/resize.spec.ts
```diff
@@ -545,6 +523,7 @@ describe('withoutReduction', () => {
       //@ts-expect-error we know this is safe
       const { image } = await applyTransforms([resize({ withoutReduction: 'true', width: '300;900' }, dirCtx)], img)
 
+      expect(dirCtx.warn).toHaveBeenCalledOnce()
```
tlaundal, line 526: Not sure we want these assertions in the existing tests, but they have been added in one (out of two) tests where there was a warning, and in one test where there was no warning.

packages/rollup/src/__tests__/main.test.ts
```diff
@@ -159,9 +159,42 @@ describe('rollup-plugin-imagetools', () => {
...
+                      return (image) => image
+                    }
+                  ]
+                }
+              })
+            ]
+          })
+
+          if (shouldPrintWarning) {
+            expect(spy.mock.calls[0][0].message).toContain('A warning')
```
tlaundal, line 192: The message logged by vite includes color codes if the terminal supports it.

```ts
expect(spy).toHaveBeenCalledWith("[plugin:imagetools] A warning")
```
Only worked if I ran `npm run test | cat` so as to disable colors. No color codes are inserted inside the warning text, so we can find it with `expect(...).toContain`.

packages/rollup/src/index.ts
```diff
@@ -71,10 +68,13 @@ export function imagetools(userOptions: Partial<RollupPluginOptions> = {}): Plug
 
       for (const config of imageConfigs) {
         const { transforms, warnings } = generateTransforms(config, transformFactories)
-        warnings.forEach((warning) => this.warn(warning))
```
tlaundal, line 74: Logging the warnings has to happen after `applyTransforms`, because this is where they are triggered.

packages/core/src/transforms/resize.ts
```diff
@@ -41,7 +41,7 @@ function parseAspect(aspect: string): number | undefined {
   return aspectRatio
 }
 
-export const resize: TransformFactory<ResizeOptions> = (config) => {
+export const resize: TransformFactory<ResizeOptions> = (config, context) => {
```
benmccann, line 44: what is context and why is `resize` the only transform being changed?

tlaundal: `resize` is the only transform that currently issues warnings.

`context` was some existing boilerplate already included for this purpose it seems. Implementing this functionality definitively gave me the impression that someone had already planned it to be implemented exactly like this. The `context` is created [here](https://github.com/JonasKruckenberg/imagetools/blob/main/packages/core/src/lib/generate-transforms.ts#L8).

benmccann:
I noticed prettier was reformatting some files in your PR. The whole repo should have been formatted, but the enforcement for it accidentally got turned off. I re-enabled it and formatted the repo, so you'll have to rebase this PR now

packages/core/src/transforms/resize.ts
```diff
@@ -91,8 +91,8 @@ export const resize: TransformFactory<ResizeOptions> = (config) => {
       finalWidth = originalWidth
       finalAspect = originalAspect
 
-      console.warn(
-        '[vite-imagetools] withoutEnlargement or withoutReduction enabled. Image width, height and aspect ratio reverted to original values'
+      context.warn(
```
benmccann, line 94: This warning doesn't seem very useful. E.g. it doesn't tell you what image it encountered

I also wonder if this should even be a warning? This doesn't seem like an error condition, but just regular functioning. Maybe it should be something more like `debug` or `info`. I'm not that familiar with the feature though. @sawyerclick can you provide some more details?

benmccann (commented):
I'm not entirely sure we need a new option as much as just lowering the level of the newly added log message. If we were to add a new option maybe it'd make sense to align it with Vite? https://vitejs.dev/config/shared-options.html#loglevel

packages/core/src/transforms/resize.ts
sawyerclick: @JonasKruckenberg recommended the warning as a part of his review in the original pr #461

Now that I'm looking at it again, I don't believe the warning is needed. It's redundant as it's sole purpose is to indicate that the flags are working.

benmccann: Ah, I see: https://github.com/JonasKruckenberg/imagetools/pull/461#discussion_r1060480418

@JonasKruckenberg it seems this warning has been a bit noisy. Do you mind if we either remove it or demote it to an `info` or `debug`? I think if users need to figure out whether it was triggered they can always compare the dimensions on the filesystem to the rendered dimensions

JonasKruckenberg: My concern was that the whole idea of `without*` is to not do anything when the conditions are met. To the uninitiated (or tired) this might seem like the whole plugin not doing it's job though. Also in all cases I can imagine `without*` is a safeguard that trips when you made an error while specifying your image dimensions. That's why I originally recommended the warning, but I guess `info` makes sense too (since it technically works as intended)

JonasKruckenberg:
> I'm not entirely sure we need a new option as much as just lowering the level of the newly added log message. If we were to add a new option maybe it'd make sense to align it with Vite? https://vitejs.dev/config/shared-options.html#loglevel

I would like to add more logs to this plugin to make issues easier to debug, but the specifics really don't matter here. Respecting the vite config value (when we're in the vite plugin) makes total sense to me, but we need a way to pass this on to the core crate. Maybe we can use an existing logger library

tlaundal:
I won't get into the discussion of whether to log when `without*` takes effect, as long as it can be disabled.

Looking into the logging capabilities of vite and rollup, I have found that:
 - Rollup has [`warn`](https://rollupjs.org/guide/en/#thiswarn) and [`error`](https://rollupjs.org/guide/en/#thiserror) functions, but nothing else it seems. The only difference is that `error` will abort bundling
 - Vite, of course, inherits these functions, but also introduces the [`config.logger`](https://vitejs.dev/guide/api-javascript.html#resolvedconfig) property to access a [`Logger`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/logger.ts) object, which supports `info`. Btw, in Vite, the plugin context `.error` aborts bundling same as in rollup, but the  `logger.error` function does not

Based on this, I think it makes sense that the imagetools rollup and vite plugins provides an object with `info`, `warn` and `error` functions to the transformer context. In this case, vite can control the logging verbosity, but I don't think rollup provides such controls.

JonasKruckenberg:
> Based on this, I think it makes sense that the imagetools rollup and vite plugins provides an object with info, warn and error functions to the transformer context. In this case, vite can control the logging verbosity, but I don't think rollup provides such controls.

I think this makes sense, when used in a Vite context we can just proxy the log functions provided by Vite and in a the rollup case we just use the functions we're provided and maybe make the rest no-ops

tlaundal:
> in the rollup case we just use the functions we're provided and maybe make the rest no-ops

Yes, no-op or `console.log` are both valid choices I think.
One thing we should clarify is whether our `error` should abort bundling, like the [plugin context](https://rollupjs.org/guide/en/#plugin-context) error function, or just log the message, like what the vite logger does. (Vite also aborts in the case of the plugin context error function).

I can implement this tomorrow if there are no concerns from Ben  or others.

docs/README.md
```diff
@@ -1,5 +1,7 @@
 imagetools / [Modules](modules.md)
 
+imagetools / [Modules](modules.md)
+
```
benmccann, line 4: we should manually remove this until https://github.com/tgreyuk/typedoc-plugin-markdown/issues/385 is fixed
```suggestion
```

.changeset/bright-bears-bake.md
```diff
@@ -0,0 +1,8 @@
+---
+'imagetools-core': minor
+'rollup-plugin-imagetools': minor
+'vite-imagetools': minor
+---
+
+Log messages through Vite and Rollup. Log level can be controlled in Vite and Rollup through the `logLevel` and `onwarn`
+options respectively.
```
benmccann, line 8: there should not be a line break in the changeset
```suggestion
feat: log messages through Vite and Rollup. The log level can be controlled in Vite and Rollup through the `logLevel` and `onwarn` options respectively.
```

benmccann (approved):

.changeset/bright-bears-bake.md
tlaundal: Seems the linebreak is needed to satisfy prettier?

benmccann: Ah, I don't think I've run into having a long message before :smile: How about we just shorten it up?
```suggestion
feat: log messages through Vite and Rollup. Allows the log level to be set with `logLevel` and `onwarn` respectively.
```

JonasKruckenberg (approved):
Looks good! Thanks for taking the time to fix this 👍🏻 One last thing though: I would argue this should be a `patch` fix too, since it doesn't introduce any new user-facing features, and instead fixes some weird behaviour

tlaundal:
> Looks good! Thanks for taking the time to fix this 👍🏻 One last thing though: I would argue this should be a `patch` fix too, since it doesn't introduce any new user-facing features, and instead fixes some weird behaviour

Haha, I originally had it as a major (because I completely removed the config option at first, instead of deprecating), so now we've been through the whole spectrum. I'll make the change to patch.

And thank you for creating this great library in the first place :raised_hands:

PR #489, opened 2023-01-17
support Vite's `base` config when using the dev server

### LINKED ISSUE #396: Images don't show in dev mode with Laravel backend
If you try this plugin with a Laravel + Vite + Vue stack, all the images will throw a `404 (Not Found)` error and not show on the page.

This is because the tool is returning relative paths and normally an asset URL imported through Vite would have an absolute URL, and in the context of a backend such as Laravel the lack of absolute URL breaks the images.

Assuming you're running in `dev` mode
Without `imagetools` loader ( normal image importing ):
```ts
import Cat from '@/images/cat.png'

console.log(Cat)
// Output: http://localhost:3000/resources/images/cat.png
```
With `imagetools` loader:
```ts
import Cat from '@/images/cat.png?width=200'

console.log(Cat)
// Output: /@imagetools/5812d5ee51f96cb9c239acca9f2ecc22bd83998a
// Throws a not found error.
```

Note that `build` works fine. It's only a `dev` mode issue.

DeedleFake:
- **Quick Checklist**

It adds support for Vite's `base` config in the dev server. This fixes issues with serving assets through imagetools when using a custom backend, such as Rails.

Fixes #396.

The URL prefix used to watch for imagetools requests is now `${viteConfig.base}/@imagetools/` instead of just `/@imagetools/`. This allows for things like Rails to properly forward those requests to the Vite dev server.

  PR?)

I don't think so, no.

### DIFF
--- a/.changeset/pink-yaks-jam.md
+++ b/.changeset/pink-yaks-jam.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': minor
+---
+
+support Vite's base config when the dev server is running
--- a/.changeset/pink-yaks-jam.md
+++ b/.changeset/pink-yaks-jam.md
@@ -1,5 +1,5 @@
 ---
-'vite-imagetools': minor
+'vite-imagetools': patch
 ---
 
 support Vite's base config when the dev server is running
--- a/.changeset/pink-yaks-jam.md
+++ b/.changeset/pink-yaks-jam.md
@@ -2,4 +2,4 @@
 'vite-imagetools': patch
 ---
 
-support Vite's base config when the dev server is running
+fix: support Vite's `base` config
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -46,7 +46,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
     enforce: 'pre',
     configResolved(cfg) {
       viteConfig = cfg
-      basePath = viteConfig.base || '/'
+      basePath = viteConfig.base?.replace(/\/$/, '') || ''
     },
     async load(id) {
       if (!filter(id)) return null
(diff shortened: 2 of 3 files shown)


packages/vite/src/index.ts
```diff
@@ -118,8 +120,8 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
 
     configureServer(server) {
       server.middlewares.use((req, res, next) => {
-        if (req.url?.startsWith('/@imagetools/')) {
-          const [, id] = req.url.split('/@imagetools/')
+        if (req.url?.startsWith(`${basePath}@imagetools/`)) {
```
benmccann, line 123: This needs to be adjusted because `basePath` may or may not end with `/`

benmccann (approved):

.changeset/pink-yaks-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': minor
```
benmccann, line 2: @JonasKruckenberg do you have any opinion about whether this fix should be `patch` or `minor`?

DeedleFake: I wasn't sure, either. It's technically a bug fix but it also technically adds functionality.

DeedleFake:
Is there anything else you'd like me to do before this can be merged? My current workaround involves manually building `vite-imagetools` and then repackaging as a minimal local package so that I can use the fixed version, and I'd very much prefer to be able to use the actual package itself.

.changeset/pink-yaks-jam.md
JonasKruckenberg: It doesn’t really matter, but I feel `patch` is most appropriate bc it’s about fixing compatibility with frameworks right?

JonasKruckenberg (approved):

.changeset/pink-yaks-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+support Vite's base config when the dev server is running
```
benmccann, line 5: I think this would also affect `build` and not just `dev` since it changes code in `load`
```suggestion
fix: support Vite's `base` config
```

DeedleFake: `build` seems to work the same for me both before and after adding `base` support.

Issue #510, opened 2023-02-14
Unexpected metadata results after image import & transformation


nick-vincent:
Hello there! 👋

I've set up a [barebones SvelteKit repo](https://github.com/nick-vincent/vite-imagetools-metadata-example) to demonstrate some unexpected metadata results using `vite-imagetools`.

In case you have trouble running it, here is a screenshot from the demo:

<img width="589" alt="image" src="https://user-images.githubusercontent.com/31055081/218856793-235ea886-c9be-469d-8972-0ea79a48d880.png">

Basically, I'm using `vite-imagetools` to import a 1600x900 PNG image and resize it to fit inside 480x480.

The image is correctly converted and resized! However, the metadata returned is not what I'd expect:

```
Image.aspect: undefined (should be 1.778 or 16:9)
Image.height: 480 (should be 270)
Image.width: 480 (should be 480)
```

As you can see, the `aspect` ratio is not returned at all. Additionally, I'd expect the `height` and `width` metadata to reflect the final dimensions of the processed image, not the `h` & `w` values passed via the query param.

Is this user error on my part, or is this returning incorrect metadata? And am I posting this issue in the right place? 🤔 

Thanks in advance for looking into this, and let me know if i can help further diagnose!

nick-vincent:
Hi again @JonasKruckenberg!

I was looking at [this code](https://github.com/JonasKruckenberg/imagetools/blob/main/packages/core/src/transforms/resize.ts#L99-L101) and it does seem like the intention is to return the final metadata values of the transformed image:

```js
    setMetadata(image, 'height', finalHeight)
    setMetadata(image, 'width', finalWidth)
    setMetadata(image, 'aspect', finalAspect)
```

I tried downgrading `imagetools-core` to v3.2.3 which does fix the aspect ratio (but not the computed width & height): 

<img width="605" alt="image" src="https://user-images.githubusercontent.com/31055081/229323423-b10f04d7-3e0f-46ef-a0fb-b38f6a591878.png">

So it seems something broke in v3.3.0, maybe in @sawyerclick's [PR](https://github.com/JonasKruckenberg/imagetools/pull/461) to add `withoutEnlargment` & `withoutReduction`?

sawyerclick:
Thanks for flagging! I'll check out the aspect ratio issue promptly.

On the width and height, are you certain those stem from the withoutEnlargement and withoutReduction additions? If so, I can investigate.

nick-vincent:
[...]

PR #524, opened 2023-04-02
fix: correctly calculate resize metadata when given w & h

nick-vincent:
Meant to fix https://github.com/JonasKruckenberg/imagetools/issues/510

There is an additional case not being handled in [`resizeTransform`](https://github.com/JonasKruckenberg/imagetools/blob/main/packages/core/src/transforms/resize.ts#L58): If both `width` AND `height` are given, but no `aspect`.

This case is a bit complicated, because the `fit` needs to be considered and possibly used to recalculate all the values:

api-resize-fit

(See [Resizing images](https://sharp.pixelplumbing.com/api-resize) in the Sharp docs)

I couldn't find anything in [resize.spec.ts](https://github.com/JonasKruckenberg/imagetools/blob/main/packages/core/src/transforms/__tests__/resize.spec.ts) testing the metadata, any pointers there?

---

**Quick Checklist**


packages/core/src/transforms/resize.ts
```diff
@@ -96,18 +123,21 @@ export const resize: TransformFactory<ResizeOptions> = (config, context) => {
...
 
     return image.resize({
-      width: Math.round(finalWidth) || undefined,
-      height: Math.round(finalHeight) || undefined,
+      width: finalWidth || undefined,
+      height: finalHeight || undefined,
       withoutEnlargement: withoutEnlargement,
       withoutReduction: withoutReduction,
-      fit: getFit(config, image),
+      fit: fit,
```
benmccann, line 140: this can be written as:
```suggestion
      fit,
```

benmccann:
For testing, I wonder if you could refactor out [the code for calculating dimensions](https://github.com/JonasKruckenberg/imagetools/blob/e538705a310633be2e78c77e10094a6fb1f7897f/packages/core/src/transforms/resize.ts#L59-L127) and then just test that part with a unit test.

benmccann:
@nick-vincent this PR will need a rebase

benmccann:
I think I found some additional edge cases, so I went ahead and merged this so that I can add tests for all of them together at once

nick-vincent:
Thanks for merging this, @benmccann! Feel free to tag me in the follow-up PR.

PR #529, opened 2023-04-10
Remove rollup-plugin-imagetools

benmccann:
Rollup is great for building libraries, but Vite is better for building apps, which is where you'd use an image plugin so basically no one uses `rollup-plugin-imagetools`. I'd like to propose dropping it in order to simplify maintenance of this library

https://npmcharts.com/compare/vite-imagetools,rollup-plugin-imagetools?log=false&interval=7

Screenshot from 2023-04-09 19-21-03


JonasKruckenberg:
I think we should publish one last release of the package with a readme update first. But in general: agreed 👍

benmccann:
> I think we should publish one last release of the package with a readme update first. 

Good idea. Here's a PR for that: https://github.com/JonasKruckenberg/imagetools/pull/530

PR #540, opened 2023-04-27
feat: img format

benmccann:
- **Quick Checklist**

* [X] I have written new tests, as applicable (for bug fixes / features)
* [X] I have added a changeset, if applicable

- **What kind of change does this PR introduce?** feature

- **What is the new behavior (if this is a feature change)?** A format for building `img` tags. Helpful when you are targeting only a single format, but potentially multiple sizes. E.g. serve only `webp` to everyone, but in multiple `sizes` using a `srcset`

- **Does this PR introduce a breaking change?** No

- **Other information**: @dummdidumm is using this in a Svelte image component that he's building


packages/core/src/output-formats.ts
```diff
@@ -31,6 +31,38 @@ export const sourceFormat: OutputFormat = () => (metadatas) => {
...
+    w: largestImage?.width as number,
+    h: largestImage?.height as number
+  }
+
+  if (metadatas.length >= 2) {
+    result.srcset = []
+    for (let i = 0; i < metadatas.length; i++) {
+      const m = metadatas[i]
+      // we don't need to create a srcset if there is only a single image
+      if (metadatas.length < 2) {
```
dummdidumm, line 56: `if (metadatas.length >= 2)` above makes this check unnecessary

packages/core/src/output-formats.ts
```diff
@@ -31,6 +31,38 @@ export const sourceFormat: OutputFormat = () => (metadatas) => {
...
 }
 
+export const imgFormat: OutputFormat = () => (metadatas) => {
+  let largestImage
+  let largestImageSize = 0
+  for (let i = 0; i < metadatas.length; i++) {
+    const m = metadatas[i]
+    if ((m.width as number) > largestImageSize) {
+      largestImage = m
+      largestImageSize = m.width as number
```
dummdidumm, line 41: I'm wondering (and whether that's actually a good idea) is if it's possible 
- to use the original image size as the one for `src/w/h`. Does that make sense? Since upscaling is lossy
- to cap the generation of images (and output meta) to "not more than 2 times the original size" because upscaling beyond that is probably wasteful (?)

dummdidumm (commented):
Is the idea to in `@sveltejs/image` set `defaultDirectives` to set `as=img` and `width=<what we have configured>`? Is it possible for the user to overwrite the default `width` if they want to (and the default is not applied then)?

packages/core/src/output-formats.ts
benmccann: oops. yeah, it started there and I meant to move it out, but guess I only got the job half done :smile: thanks for catching!

packages/core/src/output-formats.ts
benmccann: I was assuming that we would never upscale images. I just did a quick search and https://imagekit.io/blog/how-to-resize-image-in-html/ suggests the same

benmccann: > I was assuming that we would never upscale images.

Though I guess that needs to happen in `imagetools` because `@sveltejs/image` doesn't know the size of the original image. I'll check that and implement as a separate PR if it's not already happening

benmccann:
> Is the idea to in @sveltejs/image set defaultDirectives to set as=img and width=<what we have configured>? Is it possible for the user to overwrite the default width if they want to (and the default is not applied then)?

Yes, I hope that would work well

flevi29:
@benmccann Was there any documentation added for this? [I can't find it here.](https://github.com/JonasKruckenberg/imagetools/blob/main/docs/_media/directives.md#metadata)

The only reason I stumbled on this is because I was looking through the type declaration looking for a similar feature.

PR #587, opened 2023-07-31
fix: multiple transforms now survive `resolveConfig`

### LINKED ISSUE #585: imagetools-core 4.0.4 `resolveConfigs` behavior change, `format` directive no longer working
Hi, in 4.0.3:
```ts
resolveConfigs([ [ 'w', [ '5', '10' ] ], [ 'format', [ 'webp', 'avif', 'png' ] ] ], {})
```
equals
```ts
[
  { w: '5', format: 'webp' },
  { w: '5', format: 'avif' },
  { w: '5', format: 'png' },
  { w: '10', format: 'webp' },
  { w: '10', format: 'avif' },
  { w: '10', format: 'png' }
]
```

However, in 4.0.4, format is no longer parsing:
```ts
[
  { w: '5' },
  { w: '5' },
  { w: '5' },
  { w: '10' },
  { w: '10' },
  { w: '10' }
]
```

I don't see any notes on format's behavior changing. I believe this diff might be the cause of this new behavior:
https://github.com/JonasKruckenberg/imagetools/pull/583/files#diff-eb99c1eca6cc7f63c9eac72310407235a40762ebe6169de94d2b79102afd7d3e

DylanLacey:
When providing multiple transforms (eg `width` and `format`), only the initial transform was being applied.

This was due to `packages/core/src/lib/resolve-configs.ts` calling `Object.fromEntries` with an array containing all option/option names, instead of an array of arrays of key/value pairs (See [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) for the method documentation).

This commit does some massaging, using the `options` array's Iterator directly so we can grab each pair.  The `for` loop is a little ... _traditional_ ... But I implemented a `while` loop checking for Iterator doneness first and it was worse.

(This code assumes there will always be an even number of entries in `options`, mainly because if there aren't, `Object.fromEntries` won't return anything sensible anyway.)

fixes #585.

- **Quick Checklist**

* [ X ] I have read [the contributing guidelines](../CONTRIBUTING.md)
* [ X ] I have written new tests, as applicable (for bug fixes / features)
* [ X ] Docs have been added / updated (for bug fixes / features)
* [ N/A ] I have added a changeset, if applicable

Bug fix

N/A

No

### DIFF
--- a/packages/core/src/__tests__/generate-configs.spec.ts
+++ b/packages/core/src/__tests__/generate-configs.spec.ts
@@ -52,6 +52,39 @@ describe('generateConfigs', () => {
     expect(res).toHaveLength(2)
   })
 
+  // test('the returned array length equals the product of all arguments', () => {
+  //   {
+  //     const e: [string, string[]][] = [['width', ['300', '400']]]
+
+  //     const res = resolveConfigs(e, builtinOutputFormats)
+
+  //     expect(res).toHaveLength(2)
+  //   }
+  //   {
+  //     const e: [string, string[]][] = [
+  //       ['width', ['300', '400']],
+  //       ['test', ['foo', 'bar']]
+  //     ]
+
+  //     const res = resolveConfigs(e, builtinOutputFormats)
+
+  //     console.log(res)
+
+  //     expect(res).toHaveLength(4)
+  //   }
+  //   {
+  //     const e: [string, string[]][] = [
+  //       ['width', ['300', '400']],
+  //       ['test', ['foo', 'bar']],
+  //       ['height', ['100', '700']]
+  //     ]
+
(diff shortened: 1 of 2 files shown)


nizos:
Thanks for creating this PR. I was facing the same issue myself. Just a note, wouldn't it be better to not have commented out tests and console logs in the pushed code?

benmccann:
thank you! I'm using your test to verify another fix in https://github.com/JonasKruckenberg/imagetools/pull/588. I'll close this in favor of that PR and cut a new release as soon as the CI finishes running

DylanLacey:
@nizos It absolutely would; I just missed those before committing >.>

@benmccann No worries!  Your fix is much nicer ;)

PR #605, opened 2023-09-10
chore: generate image IDs from relative filepath

liegeandlief:
- **Quick Checklist**

Bug fix in order to ensure that generated image IDs are consistent across environments.

Image IDs are now generated from a relative file path instead of an absolute one which should ensure their consistency across environments.

  PR?)

Yes. Generated image IDs will change so tests which rely on these IDs may break.

- **Other information**:

### DIFF
--- a/.changeset/tender-jobs-double.md
+++ b/.changeset/tender-jobs-double.md
@@ -2,4 +2,4 @@
 'imagetools-core': patch
 ---
 
-Image IDs are now generated from relative filepaths instead of absolute ones.
+chore: image IDs are now generated from relative filepaths instead of absolute ones
--- a/.changeset/tender-jobs-double.md
+++ b/.changeset/tender-jobs-double.md
@@ -1,0 +1,5 @@
+---
+'imagetools-core': major
+---
+
+Image IDs are now generated from relative filepaths instead of absolute ones.
--- a/.changeset/tender-jobs-double.md
+++ b/.changeset/tender-jobs-double.md
@@ -1,5 +1,5 @@
 ---
-'imagetools-core': major
+'imagetools-core': patch
 ---
 
 Image IDs are now generated from relative filepaths instead of absolute ones.
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -1,7 +1,7 @@
 import { createHash } from 'node:crypto'
+import path from 'node:path'
 import sharp from 'sharp'
 import { ImageConfig } from './types.js'
-import path from 'path'
 
 export function loadImage(path: string) {
   return sharp(path)
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -9,6 +9,7 @@ export function loadImage(path: string) {
 
 export function generateImageID(url: URL, config: ImageConfig) {
   // this isn't a valid URL, but just a string used for an identifier
+  // use a relative path in the local case so that it's consistent across machines
   const baseURL = url.host
     ? new URL(url.origin + url.pathname)
     : new URL(url.protocol + path.relative(process.cwd(), url.pathname))
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -10,6 +10,7 @@ export function loadImage(path: string) {
 export function generateImageID(url: URL, config: ImageConfig) {
   const baseURL = url.host
     ? new URL(url.origin + url.pathname)
+    // use a relative path here so that the path is the same locally and on CI
     : new URL(url.protocol + path.relative(process.cwd(), url.pathname))
 
   return createHash('sha1').update(baseURL.href).update(JSON.stringify(config)).digest('hex')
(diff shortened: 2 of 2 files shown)


packages/core/src/util.ts
```diff
@@ -7,7 +8,9 @@ export function loadImage(path: string) {
 }
 
 export function generateImageID(url: URL, config: ImageConfig) {
-  const baseURL = url.host ? new URL(url.origin + url.pathname) : new URL(url.protocol + url.pathname)
+  const baseURL = url.host
+    ? new URL(url.origin + url.pathname)
+    : new URL(url.protocol + path.relative(process.cwd(), url.pathname))
```
benmccann, line 13: i'm confused as to what this line is doing. I think `url.protocol` is something like `http:` and `path.relative` will return something like `static/image.png` and thus you'll end up with a URL like `https:static/image.png` which would be an invalid URL, right?

.changeset/tender-jobs-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'imagetools-core': major
```
benmccann, line 2: why wouldn't this just be a bug fix? the API doesn't change at all
```suggestion
'imagetools-core': patch
```

packages/core/src/util.ts
liegeandlief: It would be an invalid URL but I don't think it matters. It just needs to be a string which is unique for each image which is then hashed to generate an ID string. The change here is that instead of hashing a string containing an absolute file path to the image, we're hashing a string containing a relative path to the image.

.changeset/tender-jobs-double.md
liegeandlief: It's a breaking change in as much as it will change the image IDs produced and these image IDs may be being used in various ways by consumers of the package. For instance this change would break some snapshot tests in my project where image IDs in the snapshot would no longer match and I would need to update the snapshots to contain the new IDs.

liegeandlief:
@benmccann I replied to your comments, are you able to take a look at my replies? Thanks :)

.changeset/tender-jobs-double.md
benmccann: given that it only affects tests and not actual users I'm going to go ahead and say it's not worth a major version bump. if we're too strict we're going to end up on version 100 before long

.changeset/tender-jobs-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'imagetools-core': patch
+---
+
+Image IDs are now generated from relative filepaths instead of absolute ones.
```
benmccann, line 5:
```suggestion
chore: image IDs are now generated from relative filepaths instead of absolute ones
```

packages/core/src/util.ts
benmccann, line 13:
```suggestion
    // use a relative path here so that the path is the same locally and on CI
    : new URL(url.protocol + path.relative(process.cwd(), url.pathname))
```

packages/core/src/util.ts
```diff
@@ -7,7 +8,10 @@ export function loadImage(path: string) {
 }
 
 export function generateImageID(url: URL, config: ImageConfig) {
-  const baseURL = url.host ? new URL(url.origin + url.pathname) : new URL(url.protocol + url.pathname)
+  const baseURL = url.host
```
benmccann, line 11:
```suggestion
  // this isn't a valid URL, but just a string used for an identifier
  const baseURL = url.host
```

benmccann (approved):

packages/core/src/util.ts
```diff
@@ -7,7 +8,11 @@ export function loadImage(path: string) {
 }
 
 export function generateImageID(url: URL, config: ImageConfig) {
-  const baseURL = url.host ? new URL(url.origin + url.pathname) : new URL(url.protocol + url.pathname)
+  // this isn't a valid URL, but just a string used for an identifier
+  const baseURL = url.host
+    ? new URL(url.origin + url.pathname)
+    // use a relative path here so that the path is the same locally and on CI
```
benmccann, line 14:
```suggestion
```

packages/core/src/util.ts
```diff
@@ -7,7 +8,10 @@ export function loadImage(path: string) {
 }
 
 export function generateImageID(url: URL, config: ImageConfig) {
-  const baseURL = url.host ? new URL(url.origin + url.pathname) : new URL(url.protocol + url.pathname)
+  // this isn't a valid URL, but just a string used for an identifier
+  const baseURL = url.host
```
benmccann, line 12:
```suggestion
  // use a relative path in the local case so that it's consistent across machines
  const baseURL = url.host
```

benmccann:
@liegeandlief do you use `imagetools-core` directly in your project or do you use it via `imagetools-vite`? This function has another bug and I may move it to `imagetools-vite` when fixing it so that we're more free to change it rather than leaving it exposed

Issue #86, opened 2021-05-03
Add a low-quality-placeholder directive


JonasKruckenberg:
This has been a requested feature and one that I'd argue is a wothwhile addition to the library.
I'm not quite sure on the specific syntax though and would like to get some feedback!

My proposal would be:
- **Keyword**: `lqip`
- **Type**: _boolean_

Where lqip invokes the following steps:
1. resize to a width of 640px 
2. set quality to 50

The question is wether these values should be customizeable on if so how? An integer that scales the values accordingly?
I don't want this directive to do too much since stuff like blurring can and should be added by the user to their liking.

_Originally posted by @rchrdnsh in https://github.com/JonasKruckenberg/imagetools/discussions/69_

cryptodeal:
I agree on blurring, that should be handled by the user to fit their desired aesthetic. 

I do think both the size and quality should be customizable to an extent (I like the idea of using an integer that scales the values accordingly), but I don't think there's really a need for fine grained controls given lqip would be, ultimately, for the sake of convenience. 

If a user needs the ability to highly customize the placeholder image, they should import the placeholder using the provided directives as that's going to give the most granular control in terms of the resulting image. 

E.g. I currently import placeholder images, which is passed into a lazy loading responsive image component (along w the srcset imported separately), by doing the following: 
```
import logo from './logo.jpg?w=300&blur=100&quality=30';
```

michael0liver:
+1 I would love this feature.

Great library btw it works great!

michael0liver:
> E.g. I currently import placeholder images, which is passed into a lazy loading responsive image component (along w the srcset imported separately), by doing the following:



Hey @cryptodeal do you have an example of that component?

michael0liver:
There is another style of placeholder which I really like "traced placeholder". It is included with the [`gatsby-plugin-image`](https://github.com/gatsbyjs/gatsby/issues/2435) and [`svelte-image`](https://github.com/matyunya/svelte-image) components.

I created a custom directive using the `svelte-image` [implementation](https://github.com/matyunya/svelte-image/blob/e41e7b5960f4f108094c3877859b6257b97c2e7a/src/main.js#L165-L187) as a guide:
```js
import { imagetools } from "vite-imagetools";
import { setMetadata } from "imagetools-core";

import potrace from "potrace";
import { promisify } from "util";
import { optimize } from "svgo";

const trace = promisify(potrace.trace);

function svgPlaceholderTransform(config) {
[...]

PR #662, opened 2023-11-29
feat: add picture-lqip format with low quality base64 image

pzerelles:
- **Quick Checklist**

Feature

The picture output can have a low quality base64 image.

  PR?)
No

- **Other information**:

### DIFF
--- a/.changeset/eight-chicken-worry.md
+++ b/.changeset/eight-chicken-worry.md
@@ -1,0 +1,5 @@
+---
+'imagetools-core': patch
+---
+
+feat: add picture-lqip format with low quality base64 image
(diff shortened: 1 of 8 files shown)


benmccann:
If I understood the code here correctly, it looks like this isn't really transforming the image in a way that makes it much smaller (e.g. to display just the dominant background color). I think if we were going to add a feature like this we should do it with different transforms that shrink the image in various ways as mentioned in https://github.com/JonasKruckenberg/imagetools/issues/86#issuecomment-1765384288. We don't have to add them all off the bat, but it might be nice to have two options people could choose between to help us figure out the API

pzerelles:
It actually takes the smallest size that you specify in the transform for the low quality preview.

PR #672, opened 2023-12-10
feat: add "inline" directive

### LINKED ISSUE #253: Add support for inlined images by respecting the `?inline` suffix
Currently, we can't import inlined images using the `?inline` suffix. But it's a necessary feature to have and also the current behavior is breaking a native functionality. So, I think it should be implemented.

pzerelles:
- **Quick Checklist**

feature

images can be inlined with the `inline` directive

  PR?)
no

- **Other information**:
fixes #253


packages/vite/src/index.ts
```diff
@@ -1,4 +1,5 @@
 import { basename, extname } from 'node:path'
+import { readFile } from 'node:fs/promises'
```
benmccann, line 2:
```suggestion
```

pzerelles: Corrected and rebased.

benmccann:
This one has a merge conflict I think from merging your other PR. Can you update it? Thanks!

packages/vite/src/index.ts
```diff
@@ -126,18 +126,22 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
...
+          metadata.src = `data:image/${metadata.format};base64,${(await image.toBuffer()).toString('base64')}`
         } else {
-          const fileHandle = this.emitFile({
-            name: basename(pathname, extname(pathname)) + `.${metadata.format}`,
-            source: await image.toBuffer(),
-            type: 'asset'
-          })
-
-          metadata.src = `__VITE_ASSET__${fileHandle}__`
+          if (viteConfig.command === 'serve') {
```
benmccann, line 132: you could probably combine this line with the previous line to make it an `else if`, which would reduce the number of lines and amount of indentation

packages/vite/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { basename, extname } from 'node:path'
+import { basename, extname, join } from 'node:path'
```
benmccann, line 1: you can't use the `join` from `path` because it will output `\` on Windows. You need the one from `path.posix` so that you always get `/`

docs/directives.md
```diff
@@ -508,3 +509,19 @@ Returns a URL to the generated image. This is the default when your directives p
...
 import Image from 'example.jpg?w=500' // the type of Image is a string and it's a URL to the transformed image
 ```
+
+---
+
+### Inline
+
+• **Keyword**: `inline`<br> • **Type**: _boolean_<br>
+
+Return base64 encoded inline image instead of URL to the generated image. This can be combined with some output directives like `url`, `srcset` or `picture`.
```
benmccann, line 519: it's pretty unclear what should happen if you use both `url` and `inline` and I'm not sure I'd recommend those two be used together
```suggestion
Return base64 encoded inline image instead of URL to the generated image. This can be combined with some output directives like `srcset` or `picture`.
```

packages/vite/src/index.ts
```diff
@@ -126,10 +126,12 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         const { transforms } = generateTransforms(config, transformFactories, srcURL.searchParams, logger)
         const { image, metadata } = await applyTransforms(transforms, img.clone(), pluginOptions.removeMetadata)
 
-        if (viteConfig.command === 'serve') {
+        if (directives.has('inline')) {
+          metadata.src = `data:image/${metadata.format};base64,${(await image.toBuffer()).toString('base64')}`
+        } else if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
```
benmccann, line 133: I think we'd still want to do this portion if `inline` and `serve` are both true so that we don't have to recompute the image on subsequent requests. maybe we could do something like:
```
let id;
if (viteConfig.command === 'serve') {
          id = await generateImageID(srcURL, config, img)
          generatedImages.set(id, image)
}
if (directives.has('inline')) {
          metadata.src = `data:image/${metadata.format};base64,${(await image.toBuffer()).toString('base64')}`
} else if (viteConfig.command === 'serve') {
         metadata.src = join(viteConfig?.server?.origin ?? '', basePath) + id
} else {
...
}
```

packages/vite/src/index.ts
pzerelles: But `import { join } from 'node:path/posix'` should be ok then?

docs/directives.md
pzerelles: Updated

packages/vite/src/index.ts
pzerelles: Updated

packages/vite/src/index.ts
benmccann: yep! looks good now. thanks

packages/vite/src/index.ts
```diff
@@ -129,7 +130,11 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = path.posix.join(viteConfig?.server?.origin ?? '', basePath) + id
+        }
+        if (directives.has('inline')) {
+          metadata.src = `data:image/${metadata.format};base64,${(await image.toBuffer()).toString('base64')}`
+        } else if (viteConfig.command === 'serve') {
+          metadata.src = join(viteConfig?.server?.origin ?? '', basePath) + id
```
benmccann, line 137: this is going to use the `id` passed into `load` rather than the image ID as it was before. we should probably rename this to `imageId` to avoid having two variables with the same name

You need to remove the `const` from line 131, make that `imageId`, and add a `let imageId;` right before the `if (viteConfig.command === 'serve') {` check. GitHub won't let me make a suggestion on those lines since they're unchanged

pzerelles: Updated

packages/vite/src/index.ts
```diff
@@ -126,10 +127,15 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         const { transforms } = generateTransforms(config, transformFactories, srcURL.searchParams, logger)
         const { image, metadata } = await applyTransforms(transforms, img.clone(), pluginOptions.removeMetadata)
 
+        let imageId = ''
```
benmccann, line 130:
```suggestion
        let imageId: string
```

.changeset/loud-comics-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': patch
```
benmccann, line 2:
```suggestion
'vite-imagetools': minor
```

benmccann (approved):

Issue #568, opened 2023-06-05
Respect Vite's "server.origin" option.


pavlo-tk:
Hello,

I noticed that while running Vite's dev server, URLs for images generated by `vite-imagetools` are relative (i.e. `/@imagetools/c224b1fb8dcc644a0adf403cda4d43b3390bbf1f`). For the most part, this is fine, but as soon as you're integrating with the backend, these relative paths break images.

Vite has `server.origin` option specifically for these cases: https://vitejs.dev/config/server-options.html#server-origin
It's also explained in the "Backend Integration" section in their docs: https://vitejs.dev/guide/backend-integration.html:

```
In order to properly serve assets, you have two options:

1. Make sure the server is configured to proxy static assets requests to the Vite server
2. Set `server.origin` so that generated asset URLs will be resolved using the back-end server URL instead of a relative path
This is needed for assets such as images to load properly.
```

Right now the only way to work with  `vite-imagetools` while the app is integrated with any backend is to rewrite URLs server-side (Apache, NGINX).

It would be nice if  `vite-imagetools` would respect `server.origin` option and we could stop hardcoding those rewrite rules.

jrmyio:
Created a PR for this:
https://github.com/JonasKruckenberg/imagetools/pull/690

PR #690, opened 2024-01-26
Add vite `origin` support for image urls in development

jrmyio:
Fixes: https://github.com/JonasKruckenberg/imagetools/issues/568

Add vite `origin` support for image urls in development.

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -128,7 +128,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = basePath + id
+          metadata.src = path.join(viteConfig?.server?.origin ?? viteConfig?.origin ?? '', basePath) + id;
         } else {
           const fileHandle = this.emitFile({
             name: basename(srcURL.pathname, extname(srcURL.pathname)) + `.${metadata.format}`,
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -128,7 +128,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = path.join(viteConfig?.server?.origin ?? '', basePath) + id;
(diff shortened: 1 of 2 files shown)


packages/vite/src/index.ts
```diff
@@ -128,7 +128,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = basePath + id
+          metadata.src = path.join(viteConfig?.server?.origin ?? viteConfig?.origin ?? '', basePath) + id;
```
benmccann, line 131: You can't specify `origin` at the top level of the Vite config, can you?
```suggestion
          metadata.src = path.join(viteConfig?.server?.origin ?? '', basePath) + id;
```

jrmyio: Looks like you are right. I think I saw it on the resolved config object but it doesn't seem to be part of the types so let's not use it in any case.

packages/vite/src/index.ts
```diff
@@ -128,7 +128,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = basePath + id
+          metadata.src = path.join(viteConfig?.server?.origin ?? '', basePath) + id;
```
benmccann, line 131: I don't think this will work on Windows and needs to be:
```suggestion
          metadata.src = path.posix.join(viteConfig?.server?.origin ?? '', basePath) + id;
```

jrmyio: Ah that makes sense!

packages/vite/src/index.ts
```diff
@@ -128,7 +128,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
         if (viteConfig.command === 'serve') {
           const id = await generateImageID(srcURL, config, img)
           generatedImages.set(id, image)
-          metadata.src = basePath + id
+          metadata.src = path.posix.join(viteConfig?.server?.origin ?? '', basePath) + id;
```
benmccann, line 131: you need to run `pnpm format`
```suggestion
          metadata.src = path.posix.join(viteConfig?.server?.origin ?? '', basePath) + id
```

benmccann (approved):

PR #708, opened 2024-03-29
fix: avif format

AdamMerrifield:
- **Quick Checklist**

Bug fix

`avif` files that are from the cache directory have the correct format

  PR?)
No

- **Other information**:
Currently when an `avif` file read from cache, Sharp metadata returns `{format: 'heif', compression: 'av1'...}` causing the file extension and Picture mime type to be incorrect.
https://github.com/lovell/sharp/issues/2504 
https://github.com/lovell/sharp/issues/3746

The new test for `cache.avifFormat` will fail on imagetools@7.0.0

### DIFF
--- a/.changeset/long-baboons-dream.md
+++ b/.changeset/long-baboons-dream.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+fix: avif format when file is from cache
--- a/.changeset/long-baboons-dream.md
+++ b/.changeset/long-baboons-dream.md
@@ -2,4 +2,4 @@
 'vite-imagetools': patch
 ---
 
-fix: avif format when file is from cache
+fix: correctly set avif format when file is read from cache
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -142,6 +142,11 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
 
         if (cacheOptions.enabled && (statSync(`${cacheOptions.dir}/${id}`, { throwIfNoEntry: false })?.size ?? 0) > 0) {
           metadata = (await sharp(`${cacheOptions.dir}/${id}`).metadata()) as ImageMetadata
+          // we set the format on the metadata during transformation using the format directive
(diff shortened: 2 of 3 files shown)


packages/vite/src/index.ts
```diff
@@ -142,6 +142,9 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
 
         if (cacheOptions.enabled && (statSync(`${cacheOptions.dir}/${id}`, { throwIfNoEntry: false })?.size ?? 0) > 0) {
           metadata = (await sharp(`${cacheOptions.dir}/${id}`).metadata()) as ImageMetadata
+          // fix avif metadata format https://github.com/lovell/sharp/issues/2504 https://github.com/lovell/sharp/issues/3746
```
benmccann, line 145: I think this doesn't explain what's happening very well. how about this:
```suggestion
          // we set the format on the metadata during transformation using the format directive
          // when restoring from the cache, we use sharp to read it from the image and that results in a different value for avif images
          // see https://github.com/lovell/sharp/issues/2504 and https://github.com/lovell/sharp/issues/3746
```

benmccann:
When restoring from the cache we only restore the image and not the metadata, which we re-read from the image. This could be an issue for other properties as well, which will end up missing when restore from the cache. Though in practice I'm not sure any properties besides `format` and `width` are used and I guess that `format` is the only one that doesn't match what is read from the image, so this fix is probably fine for most use cases. I wonder if anyone is relying on getting the directives from the metadata. Perhaps we should remove them as I don't think they need to be there and the output will be different depending on whether you restore from the cache or not

Anyway, I think this change is fine, but we should probably update the comment to better clarify exactly what is happening

AdamMerrifield:
@benmccann comment has been updated, but I agree. I'm unsure of any other formats that something similar to this would occur with.

.changeset/long-baboons-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+fix: avif format when file is from cache
```
benmccann, line 5:
```suggestion
fix: correctly set avif format when file is read from cache
```

PR #711, opened 2024-04-01
fix: Consistent image ID hashes across machines

AdrianGonz97:
This PR fixes an issue that we've been experiencing in https://github.com/huntabyte/shadcn-svelte/pull/978 where the hash for the generated image ID is different when it's computed on different machines. The goal of that PR was to cache the images so that they can be utilized across our GH runners.

The source of the issue is the `mtime` stat from the image file, which seemingly differs from machine to machine, causing the hash to generate a different id, resulting in a cache MISS. Instead, `mtime` has been replaced for the `size` stat, which (in conjunction with the image config and file url) should provide sufficient uniqueness for the hash.

---

- **Quick Checklist**

	Bug fix

	Implements consistent image ID hashing across machines	

  PR?)
	No

- **Other information**:
	I wasn't sure the best way to express this change in a test, so I've hardcoded the expected id string. I'm sure there's a better way to do it, so please feel free to modify it!

### DIFF
--- a/packages/vite/src/utils.ts
+++ b/packages/vite/src/utils.ts
@@ -1,6 +1,5 @@
 import { createHash } from 'node:crypto'
 import path from 'node:path'
-import { statSync } from 'node:fs'
 import type { ImageConfig } from 'imagetools-core'
 
 export const createBasePath = (base?: string) => {
@@ -16,8 +15,7 @@ export async function generateImageID(url: URL, config: ImageConfig, imageBuffer
   // baseURL isn't a valid URL, but just a string used for an identifier
   // use a relative path in the local case so that it's consistent across machines
   const baseURL = new URL(url.protocol + path.relative(process.cwd(), url.pathname))
-  const { size } = statSync(path.resolve(process.cwd(), decodeURIComponent(url.pathname)))
-  return hash([baseURL.href, JSON.stringify(config), size.toString()])
+  return hash([baseURL.href, JSON.stringify(config), imageBuffer])
 }
 
 function hash(keyParts: Array<string | NodeJS.ArrayBufferView>) {
--- a/packages/vite/src/utils.ts
+++ b/packages/vite/src/utils.ts
@@ -1,5 +1,6 @@
 import { createHash } from 'node:crypto'
 import path from 'node:path'
+import { statSync } from 'node:fs'
 import type { ImageConfig } from 'imagetools-core'
 
 export const createBasePath = (base?: string) => {
@@ -15,7 +16,8 @@ export async function generateImageID(url: URL, config: ImageConfig, imageBuffer
   // baseURL isn't a valid URL, but just a string used for an identifier
   // use a relative path in the local case so that it's consistent across machines
   const baseURL = new URL(url.protocol + path.relative(process.cwd(), url.pathname))
-  return hash([baseURL.href, JSON.stringify(config)])
+  const { size } = statSync(path.resolve(process.cwd(), decodeURIComponent(url.pathname)))
+  return hash([baseURL.href, JSON.stringify(config), size.toString()])
 }
 
 function hash(keyParts: Array<string | NodeJS.ArrayBufferView>) {
--- a/packages/vite/src/utils.ts
+++ b/packages/vite/src/utils.ts
@@ -6,7 +6,7 @@ export const createBasePath = (base?: string) => {
   return (base?.replace(/\/$/, '') || '') + '/@imagetools/'
 }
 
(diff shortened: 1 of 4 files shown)


benmccann:
Hmm, that's really weird. I'm surprised the mtime would be different across machines. Are they different OSes? Or is the mtime set based upon when you check the repo out from git. I'd love to better understand the issue before we switch away from it

AdrianGonz97:
> Hmm, that's really weird. I'm surprised the mtime would be different across machines. Are they different OSes? Or is the mtime set based upon when you check the repo out from git. I'd love to better understand the issue before we switch away from it

It's definitely strange 😄. The OSes are set to be the same as they're running from the same workflow, so I don't think it's that. Perhaps it's more related to timezones? 🤔

AdrianGonz97:
> Or is the mtime set based upon when you check the repo out from git

It seems like this assumption was correct, @benmccann! After checking out the repo on 2 separate instances, I'm ending up with two different `mtime` stats.

<details>
<summary>File stats 1:</summary>

```bash
{
  stats: Stats {
    dev: 2080,
    mode: 33188,
    nlink: 1,
    uid: 1000,
    gid: 1000,
    rdev: 0,
    blksize: 4096,
    ino: 768440,
    size: 159643,
    blocks: 312,
    atimeMs: 1712710759640.3115,
    mtimeMs: 1712710733650.3157,
    ctimeMs: 1712710733650.3157,
    birthtimeMs: 1712710733650.3157,
    atime: 2024-04-10T00:59:19.640Z,
    mtime: 2024-04-10T00:58:53.650Z,
    ctime: 2024-04-10T00:58:53.650Z,
    birthtime: 2024-04-10T00:58:53.650Z
  }
}
```

</details>

<details>
<summary>File stats 2:</summary>

```bash
{
  stats: Stats {
    dev: 2080,
    mode: 33188,
    nlink: 1,
    uid: 1000,
    gid: 1000,
    rdev: 0,
    blksize: 4096,
    ino: 1598355,
    size: 159643,
    blocks: 312,
    atimeMs: 1712711152420.249,
    mtimeMs: 1712711150750.2493,
    ctimeMs: 1712711150750.2493,
    birthtimeMs: 1712711150750.2493,
    atime: 2024-04-10T01:05:52.420Z,
    mtime: 2024-04-10T01:05:50.750Z,
    ctime: 2024-04-10T01:05:50.750Z,
    birthtime: 2024-04-10T01:05:50.750Z
  }
}
```

</details>

benmccann:
Sorry for the long delay. I'm afraid I'd forgotten about this PR.

I'm nervous about using the file size because you could hit false positives with it. I'm not sure how expensive it would be to compute a hash for every file, but it seems safer.

I had an idea that we could use mtime first and then fallback to the hash if it's different, but that probably wouldn't work because it would require us storing two hashes.

packages/vite/src/utils.ts
```diff
@@ -16,8 +15,7 @@ export async function generateImageID(url: URL, config: ImageConfig, imageBuffer
   // baseURL isn't a valid URL, but just a string used for an identifier
   // use a relative path in the local case so that it's consistent across machines
   const baseURL = new URL(url.protocol + path.relative(process.cwd(), url.pathname))
-  const { mtime } = statSync(path.resolve(process.cwd(), decodeURIComponent(url.pathname)))
-  return hash([baseURL.href, JSON.stringify(config), mtime.getTime().toString()])
+  return hash([baseURL.href, JSON.stringify(config), imageBuffer])
```
benmccann, line 18: hashing the image buffer has to be the most expensive part of this since it's so large relative to everything else. this has us re-calculating a hash for the image for each individual image config, which we probably don't need to do. what if we calculate the hash for the image separately first and then make the third parameter of this function be the hash of the image rather than the image itself

packages/vite/src/utils.ts
```diff
@@ -1,26 +1,24 @@
...
 import path from 'node:path'
-import { statSync } from 'node:fs'
 import type { ImageConfig } from 'imagetools-core'
 
 export const createBasePath = (base?: string) => {
   return (base?.replace(/\/$/, '') || '') + '/@imagetools/'
 }
 
-export async function generateImageID(url: URL, config: ImageConfig, imageBuffer: Buffer) {
+export function generateImageID(url: URL, config: ImageConfig, imageHash: string) {
```
benmccann, line 9: it occurs to me that we don't really need the `url` anymore now that we're no longer using the mtime. the image hash should be unique enough. I think removing it would simplify this function a fair amount

AdrianGonz97: agreed. nice catch

benmccann (approved):
thanks for all the changes and the thorough job!

PR #724, opened 2024-05-27
feat: add experimental image orientation fix

### LINKED ISSUE #700: Some images get rotated for some reason
I'm using `@sveltejs/enhanced-img`, and for some reason, my optimized images get "flipped" while the original files are the correct side up.

`+page.server.js`:
```ts
/** @type {import('./$types').PageServerLoad} */
export function load() {
	return {
		images: Object.values(import.meta.glob('$lib/assets/gallerij/*.{jpeg,png}', { query: { enhanced: true }, import: 'default', eager: true })),
		images2: Object.values(import.meta.glob('$lib/assets/gallerij/*.{jpeg,png}', { import: 'default', eager: true })),
	};
}
```
`+page.svelte`
```html
<script>
    /** @type {import('./$types').PageData} */
    export let data;
</script>
{#each data.images as image, i}
    <div class="col-12 col-md-4 col-lg-3 h-auto p-2" style="order: {2 * i}">
        <enhanced:img src={image} class="w-100 h-auto" />
    </div>
{/each}
{#each data.images2 as image, i}
    <div class="col-12 col-md-4 col-lg-3 h-auto p-2" style="order: {2 * i + 1}">
        <img src={image} class="w-100 h-auto" />
    </div>
{/each}
```

This does however not happen with all images for some reason???
Example image which gets rotated:
062nbppmfl20hed1

See below: left image is optimized, right image is original
image

happycollision:
> **NOTE:** This PR is incomplete, but the functionality is all there. It needs documentation and I need to better understand the tests in the Vite package to get those working the way I want. Everything else is ready. I don't have time for docs today, but I will get to them next week at the latest.
>
> I will also rename commit summaries to be in keeping with the format you prefer, but won't do that right away.
>
> Please feel free to start giving feedback about the runtime code and the idea itself. 

Addresses #700 via an experimental setting.

- **Quick Checklist**

🚨 I have not done everything on this list yet. I **will do them**, but I wanted to get some eyes on this draft first.

* [ ] I have added a changeset, if applicable 🚨🚨🚨 This PR needs a changeset. Just haven't done it yet. 

A new setting is added to both Vite config and as an option to pass into `applyTransforms`.

```ts
{
  experimental: {
    preserveInitialOrientation: boolean
  }
}
```

This opt-in feature will pre-`rotate`/`flip`/`flop` an image based on the EXIF orientation data so that downstream developers can think about their images without first considering whether or not the image appearance is affected by metadata or by actual pixel locations. Two images that look the same might have completely different results when applying `flip`, `flop`, or `rotate`. If you take `removeMetadata` into account, there are even more permutations to consider.

This new setting removes the need to consider all these things. 

  PR?)

I have attempted to test initial behavior and ensure that, unless you opt-in, that behavior is exactly the same. So no, there shouldn't be any breaking changes.

- **Other information**:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -32,6 +32,7 @@
     "dev": "rollup -cw",
     "build": "rollup -c",
     "test": "vitest run",
+    "test:watch": "vitest watch",
     "coverage": "vitest run --coverage"
   },
   "devDependencies": {
--- a/packages/core/src/transforms/__tests__/flip.spec.ts
+++ b/packages/core/src/transforms/__tests__/flip.spec.ts
@@ -82,11 +82,19 @@ describe('flip', () => {
       const scenarios = [
         {
           name: 'matches removeMetadata=true legacy behavior',
-          opts: { removeMetadata: true }
+          opts: { removeMetadata: true, experimental: { preserveInitialOrientation: false } }
         },
         {
           name: 'matches removeMetadata=false legacy behavior',
-          opts: { removeMetadata: false }
+          opts: { removeMetadata: false, experimental: { preserveInitialOrientation: false } }
+        },
+        {
+          name: 'matches removeMetadata=true expected behavior',
+          opts: { removeMetadata: true, experimental: { preserveInitialOrientation: true } }
+        },
+        {
+          name: 'matches removeMetadata=false expected behavior',
+          opts: { removeMetadata: false, experimental: { preserveInitialOrientation: true } }
         }
       ]
 
--- a/packages/core/src/transforms/__tests__/flip.spec.ts
+++ b/packages/core/src/transforms/__tests__/flip.spec.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { TransformFactoryContext } from '../../types'
 import { flip } from '../flip'
-import { applyTransforms } from '../../index'
+import { applyTransforms, format } from '../../index'
 import sharp, { Sharp } from 'sharp'
 import { join } from 'path'
 import { toMatchImageSnapshot } from 'jest-image-snapshot'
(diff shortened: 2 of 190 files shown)


happycollision:
I've discovered that the `.withMetadata()` method keeps all metadata while allowing you to write to some of that metadata at the same time. This means it is incompatible with the `removeMetadata` option. Looking into a different solution now...

packages/core/src/transforms/__tests__/flip.spec.ts
happycollision, line 1: Note to self: It is probably better to write tests for this experimental feature in a completely separate file, and group them all together, instead of spreading them across the existing `flip`, `flop`, and `rotate` tests.

packages/core/package.json
```diff
@@ -32,6 +32,7 @@
     "dev": "rollup -cw",
     "build": "rollup -c",
     "test": "vitest run",
+    "test:watch": "vitest watch",
```
happycollision, line 35: I couldn't get into `watch` mode without this, but I didn't mean to add it to the commit. Will drop.

packages/vite/package.json
```diff
@@ -32,6 +32,7 @@
     "dev": "rollup -cw",
     "build": "rollup -c",
     "test": "vitest run",
+    "test:watch": "vitest watch",
```
happycollision, line 35: Same here. I meant to drop this change.

packages/vite/src/types.ts
```diff
@@ -95,6 +95,22 @@ export interface VitePluginOptions {
...
   cache?: CacheOptions
+
+  /**
+   * These features are not guaranteed to work exactly the same, or even be
+   * present between minor versions.
+   */
+  experimental?: {
+    /**
+     * Sharp's handling of image orientation probably does not match the
+     * expectations of users of imagetools. Enabling this option tries tp preserve
```
happycollision, line 106: typo s/tp/to

packages/vite/src/__tests__/__image_snapshots__/main-test-ts-src-tests-main-test-ts-vite-imagetools-options-experimental-preserve-initial-orientation-true-affects-images-with-orientation-1-snap.png
happycollision, line 1: I would expect the arrow to point left in this one. I am missing something on the `vite` side of things at the moment.

happycollision:
I've upstreamed some tools into Sharp that will make this whole thing easier to accomplish.

PR #732, opened 2024-07-07
Fix url concatenation for dev mode with origin set in server options

dknibbe:
- **Quick Checklist**

This change contains a bug fix for the feature introduced in pull request #690. The server origin option usually contains a  protocol scheme identifier followed by a double slash, e.g. "http://localhost:1234". Nodes POSIX path implementation is not meant to join URLs, which means the double slash is converted to a single slash, resulting in an invalid URL, e.g "http:/localhost:1234". Using url-join instead ensures the result will be a valid URL.

- **Other information**:
Added new dependency: [url-join](https://github.com/jfromaniello/url-join), which seams to be the safest option

### DIFF
--- a/.changeset/new-otters-turn.md
+++ b/.changeset/new-otters-turn.md
@@ -2,4 +2,4 @@
 'vite-imagetools': patch
 ---
 
-Fixed url concatenation for dev mode with origin set in server options.
+fix: correct URL concatenation in dev mode when origin set in server options
--- a/.changeset/new-otters-turn.md
+++ b/.changeset/new-otters-turn.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+Fixed url concatenation for dev mode with origin set in server options.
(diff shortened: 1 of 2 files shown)


packages/vite/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { basename, extname } from 'node:path'
-import { join } from 'node:path/posix'
+import urlJoin from 'url-join'
```
benmccann, line 2: I don't think we need to add a dependency for such a simple piece of functionality. Can you just add a method here to do it instead?

dknibbe: I changed it to string concatenation. This should work fine, because Vite makes sure that the origin option never ends with a '/'.

.changeset/new-otters-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+Fixed url concatenation for dev mode with origin set in server options.
```
benmccann, line 5:
```suggestion
fix: correct URL concatenation in dev mode when origin set in server options
```

Issue #748, opened 2024-09-26
The plugin processes images with ?raw and ?url queries when it shouldn't


didoin:
I’d like to report a minor bug. When I import images using only the `?raw` or `?url` queries, the plugin processes them when it shouldn’t. As a result, the asset build generates images in `.png` format instead of preserving the original file format. It seems that if the format is not specified, the plugin defaults to converting images to `.png`.

According to the Vite documentation on [Static Asset Handling](https://vitejs.dev/guide/assets) I can import an image using these arguments to determine whether to receive the asset's URL or a string containing the image content, which is particularly useful for `.svg` files.

This issue can be easily circumvented with the following configuration:

```js
// vite.config.js
export default defineConfig({
    plugins: [
      imagetools({
        exclude: [
          'public/**/*',
          /\?(url|raw)$/, // FIX bug ?raw & ?url
        ],
    ],
});
```

However, I think it is appropriate to apply a FIX directly in the plugin code, since the `?url` and `?raw` queries are vite-specific features and should be handled automatically by the plugin.

Thank you.

benmccann:
Yeah. That sounds reasonable. PR welcome

PR #750, opened 2024-09-27
fix: exclude assets with ?raw and ?url queries from processing

didoin:
- **Quick Checklist**

This fixes the bug discussed in https://github.com/JonasKruckenberg/imagetools/issues/748.

  PR?)
No.

- **Other information**:

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -71,7 +71,7 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
       basePath = createBasePath(viteConfig.base)
     },
     async load(id) {
-      if (!filter(id)) return null
+      if (/\?(url|raw)$/.test(id) || !filter(id)) return null
 
       const srcURL = parseURL(id)
       const pathname = decodeURIComponent(srcURL.pathname)


benmccann:
I wonder if it should be something more like `\?(.*&)?(url|raw)` in case it is combined with another query parameter. Probably unlikely, but maybe safer?

didoin:
The official Vine documentation explicitly talks about `?url` suffix and `?raw` suffix.
[Static Asset Handling ](https://vitejs.dev/guide/assets#explicit-url-imports)

Also, I think your pattern doesn't allow specifying future query parameters (perhaps for new transformers) that are prefixed with `raw` or `url` (`?rawable` example,  I know it's a bad example 😂).

JonasKruckenberg:
> specifying future query parameters (perhaps for new transformers) that are prefixed with raw or url

For what its worth, I think we should not worry about that, if a situation arises in the future where there are more special query params like that we can easily push another version then.

didoin:
If you feel that a more rigorous regular expression would be more appropriate, that's perfectly fine. From my perspective, though, it might not be necessary in this case, since "vite" expects these parameters as a suffix.

As for the "future query parameters" concern, I was also considering the possibility of custom parameters that users might define through the extendTransforms function. It seems like we could avoid potential issues down the line by addressing this now.

Stadly:
This wouldn't cover the case that brought me here — the pattern is anchored, so `?url&no-inline` still goes through:

```js
import url from "./icon.png?url&no-inline"; // 2,917 B palette PNG in, 5,057 B truecolour out
```

I applied your change to the released 12.0.0: `?url` alone is skipped, `?url&no-inline` and `?raw&url` are not.

On the worry about taking away the URL of a *transformed* image — that case is `?w=200&as=url`, which this pattern doesn't match either way. Bare `url` selects nothing: the output format comes from `as` alone, and `resolveConfigs` already keeps bare `url`/`srcset` keys out of the transform combinations. So the condition could be:

skip when the query names no transform and no `as`.

That catches combined queries, leaves `?w=200&as=url` transforming, and covers `extendTransforms` directives for free.

On configurability: under `@sveltejs/enhanced-img` there is none — it builds the imagetools instance itself and takes no options, so this default is the whole behaviour for every SvelteKit user (sveltejs/kit#16923).

PR #753, opened 2024-10-09
fix: crop horizontally when aspect ratio is reduced without width or height

geigerzaehler:
When the aspect ratio is reduced and no width or height are given the image is now correctly cropped.

Before, if I requested an aspect ratio of 1:2 for an image with dimensions 640x800 it would not change the dimensions of the image. Now the image is correctly cropped to 400x800.

- **Quick Checklist**

### DIFF
--- a/packages/core/src/transforms/__tests__/resize.spec.ts
+++ b/packages/core/src/transforms/__tests__/resize.spec.ts
@@ -283,6 +283,16 @@ describe('aspect', () => {
       expect(await image.toBuffer()).toMatchImageSnapshot()
     })
 
+    test('w/ crop horizontally', async () => {
+      //@ts-expect-error we know this is safe
+      const { image } = await applyTransforms([resize({ aspect: '1:2' }, dirCtx)], img)
+
+      const { width = 0, height = 0 } = await sharp(await image.toBuffer()).metadata()
+      expect(width / height).toEqual(1 / 2)
+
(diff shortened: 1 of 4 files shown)


packages/core/src/transforms/__tests__/resize.spec.ts
```diff
@@ -283,6 +283,16 @@ describe('aspect', () => {
       expect(await image.toBuffer()).toMatchImageSnapshot()
     })
 
+    test('w/ crop horizontally', async () => {
+      //@ts-expect-error we know this is safe
```
benmccann, line 287:
```suggestion
      // @ts-expect-error we know this is safe
```

benmccann:
@geigerzaehler thanks for this. I went through all the dependency updates today and updated vitest to version 2, which changed all the snapshot files. Can you rebase this PR against the latest commit from `main`, remove your snapshop file, and generate a new one? See https://github.com/JonasKruckenberg/imagetools/pull/764

Thanks!

geigerzaehler:
> Can you rebase this PR against the latest commit from `main`, remove your snapshop file, and generate a new one? See #764

Of course, done.

packages/core/src/transforms/__tests__/resize.spec.ts
benmccann, line 287:
```suggestion
      // @ts-expect-error we know this is safe
```

benmccann (approved):

PR #754, opened 2024-10-09
fix: remove legacy import assertion unsupported by Node v22

geigerzaehler:
I wasn’t able to build the project using Node v22 because it used the legacy import assertion syntax. Support for this was removed from Node. To fix this, I’m reading `package.json` as to not rely on JSON module support which may be unavailable on older Node versions (?).

I’m also fixing the Github workflows so that they actually run against different Node versions and we can discover this in the future.

- **Quick Checklist**

* [ ] ~~Docs have been added / updated (for bug fixes / features)~~
* [ ] ~~I have added a changeset, if applicable~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,11 @@
 name: Unit tests
-on: pull_request
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


benmccann:
ah. apparently `assert` is becoming `with`: https://github.com/nodejs/node/issues/51622

PR #755, opened 2024-10-10
fix: add `originalFilename` to emitted asset to correct the `manifest.json`

riw:
- **Quick Checklist**

currently, only the last version of several conversions of an image with the same name is noted in manifest.json. this change adds an entry for each image.

  PR?)

- **Other information**:

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -1,4 +1,5 @@
 import { basename, extname } from 'node:path'
+import { relative } from 'node:path/posix'
 import { statSync, mkdirSync, createReadStream } from 'node:fs'
 import { writeFile, readFile, opendir, stat, rm } from 'node:fs/promises'
 import type { Plugin, ResolvedConfig } from 'vite'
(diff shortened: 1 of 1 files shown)


packages/vite/src/index.ts
```diff
@@ -171,7 +174,8 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
           const fileHandle = this.emitFile({
             name: basename(pathname, extname(pathname)) + `.${metadata.format}`,
             source: image ? await image.toBuffer() : await readFile(`${cacheOptions.dir}/${id}`),
-            type: 'asset'
+            type: 'asset',
+            originalFilename: relative(rootPath, id)
```
benmccann, line 178: it looks like vite also normalizes the path, so we should probably do the same:

https://github.com/vitejs/vite/blob/31cc61b7404b2b22d956856f97eeebd0a62f3b4d/packages/vite/src/node/plugins/asset.ts#L397

userquin: I guess you also need to add also `names: [basename(pathname, extname(pathname)) + `.${metadata.format}`]`, Vite 6 manifest using `asset.names.length > 0` to check if the asset with `type='assets'` should be added to the manifest:

image


`name` was deprecated time ago, Vite<6 checking if `name` is present to include the asset in the manifets:
image

benmccann:
merged in https://github.com/JonasKruckenberg/imagetools/pull/808

I tried to update this PR, but you must have unchecked the "allow edits by maintainers" box, so I had to do it as a new one

thanks for this!!

PR #795, opened 2025-04-13
Add `autoOrient` from Sharp 0.34

happycollision:
- **Quick Checklist**

@happycollision's checklist before merge

* [ ] Remove TEMP commit
* [ ] Reword CAUTION commit after getting confirmation that it is correct
* [ ] Regenerate docs again after the above commits have been altered.

This PR adds the new `autoOrient` feature from Sharp 0.34 as a directive.

This is a backwards-compatible change that adds a new directive, which needs to be explicitly invoked by downstream users if they want it.

  PR?)

It doesn't, but I believe future versions of `imagetools-vite` should apply `autoOrient` by default. (This PR doesn't do that, though.)

- **Other information**:

### State mutations

Please note that the `autoOrient` function is not pure. The [docs ask for functions without side effects](https://github.com/JonasKruckenberg/imagetools/blob/6fe10945a4ac67eb214a0e291106a0cda963f9c4/CONTRIBUTING.md?plain=1#L56), but this seems to be a guideline more than a rule, since [some existing directives also mutate state](https://github.com/JonasKruckenberg/imagetools/blob/6fe10945a4ac67eb214a0e291106a0cda963f9c4/packages/core/src/transforms/invert.ts#L12). Or perhaps the `image[METADATA]` is always fair game?

Even if altering the `[METADATA]` is considered to be an exception to keeping functions side-effect free, I still want to call out that this particular function mutates state inside `[METADATA]` that it did not create: When `autoOrient` is applied, I also change the `width` and `height` of the image metadata to reflect the re-oriented width and height via properties made available by Sharp. The reason these values need to be modified is that other directives expect that the width and height are correctly referring to the image's _final_ width and height.

### Test snapshot updates

I somewhat blindly updated the test snapshots in the "CAUTION" commit. I assumed that the reason for any deviation was simply because of a new version of Sharp. I did look at the image diffs (when present) and they sure seemed nearly the same. But I didn't spend any time trying to understand the tests and more carefully look for deviations that were unwelcome.

### TEMP commit

I added that first commit as a convenience for me since I use `asdf` for versioning `pnpm`. I will remove it before merging.

### Documentation changes and my DX suggestions

My documentation changes in the vite package suggest that almost everyone should use `autoOrient` since that produces the behavior that anyone unfamiliar with exif orientation would expect. I also suggest automatically adding that behavior in the future with a new config option to turn it off, specifically.

For now, documenting it in the defaultDirectives section seems like a decent compromise. The problem, however, is that downstream tools (such as Svelte's `enhanced-img`) might overwrite the defaults in a way that makes the end user/developer unable to automatically add the directive.

So I do suggest a follow on PR to add the default auto orient behavior as an additional vite config option.


benmccann:
Can you drop the changes regenerating the docs? The setup is pretty finicky at the moment and I'd like to do that part myself after merging this

benmccann:
I've upgraded sharp in another PR to break things up and make it easier to review your new feature. Could you please rebase?

happycollision:
@benmccann done

packages/core/src/transforms/__tests__/autoOrient.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  test('missing', () => {
+    const res = autoOrient({}, dirCtx)
+
+    expect(res).toBeUndefined()
+  })
+
+  describe('arguments', () => {
+    test('invalid', () => {
+      //@ts-expect-error invalid args
```
benmccann, line 34:
```suggestion
      // @ts-expect-error invalid args
```

benmccann:
I think we should make this the default from the beginning. We can make it a major release since it's a breaking change. The option could be called something like `noAutoOrient` or `disableAutoOrient`

happycollision:
Oh I can definitely get behind that.

happycollision:
@benmccann I've changed the PR to reflect an opt-out instead of an opt-in.

Aside from reworking the original changes from before, I also added some documentation and a new test to ensure that autoOrient runs by default.

There's something I haven't dug into yet though: I do not know the behavior of `autoOrient` when there are no other directives applied at all. I also am not sure how best to test that an image does/does not get rotated when there are no directives. I'll eventually dive in and resolve these things myself, but feel free to add your own solution before I do if you want.

sfreytag:
I've been following this PR via https://github.com/JonasKruckenberg/imagetools/issues/700 and am enthusiastic to use it. I'm maintaining a simple static site and all the image processing is automatic via imagetools, apart from me having to manually rotate images with exif orientations (typically off an iPhone). I am now using happycollision's patches and it is entirely automatic and working well. Is there anything I can do to help it over the line?

.changeset/wicked-sheep-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+'imagetools-core': major
+'vite-imagetools': major
+---
+
+Add `autoOrient` directive from Sharp 0.34.0, defaulting to automatically apply it to every image. An opt out is available via `noAutoOrient`.
```
benmccann, line 6:
```suggestion
breatking: add `autoOrient` directive from Sharp 0.34.0, defaulting to automatically apply it to every image. An opt-out is available via `noAutoOrient`.
```

PR #871, opened 2026-03-02
fix: bypass sharp pipeline for cached images

jep-a:
- **Quick Checklist**

* [x] I have read [the contributing guidelines](../CONTRIBUTING.md)
* [ ] I have written new tests, as applicable (for bug fixes / features)
  - Didn't have time to write a new test for this at the moment
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] I have added a changeset, if applicable

- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix

- **What is the new behavior (if this is a feature change)?**
Responses of cached images will be served faster

- **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this
  PR?)
Unsure at the moment

- **Other information**:

Cached images are taking as long as uncached images to serve, this is due to images being sent to sharp pipeline even if they were cached, see here for reproduction:
https://codesandbox.io/p/devbox/tn3wgp

PR #877, opened 2026-03-18
fix: serve cached images directly from disk to prevent Sharp re-encoding

sebbayer:
- **Quick Checklist**

 Bug fix: Images are being sent to the sharp pipeline even if they were cached, and output with default sharp settings. This is an alternative solution for #871. My solution also sets the correct `Content-Type` HTTP header on images in vite dev mode instead of `Content-Type: image/undefined`.

  In the original code, both the cache-hit and cache-miss paths end up with `image = sharp(cachedBuffer)` — a plain Sharp instance with no output options. When the middleware calls `image.clone().pipe(res)`, Sharp decodes the cached image to raw pixels and re-encodes it with default format settings.

  The cache file on disk is always correct. But it's never streamed directly — it always goes through a Sharp decode→re-encode round-trip that strips away any format-specific options (WebP quality, AVIF effort, etc.) or for example PNG palette/colors, when they have been set by custom transforms.

  The `!image` fallback that streams via `createReadStream` (bypassing Sharp entirely) was already there but was never reached because image was always set.

  When cache is enabled, don't store a Sharp instance in `generatedImages`. Set `image = undefined` so the existing `configureServer` middleware falls through to the `!image` branch, which streams the cache file directly from disk via `createReadStream` — no Sharp decode/re-encode involved.

  PR?)
  No

- **Other information**:


benmccann:
I just released a major update. Can you test if this is still necessary? If so, can you fix the merge conflict?

benmccann:
Would it be possible to add a test for this?

sebbayer:
@benmccann I just tested the 11.0.0 release and caching seems to be working fine now, so I think this fix is not needed anymore. I updated this merge request to contain only tests for caching.

PR #886, opened 2026-04-22
fix: use atomic write for cache files to prevent corrupt reads

nickgarlis:
- **Quick Checklist**

Node's writeFile opens the target with O_TRUNC, zeroing it before the data arrives. A concurrent build process can read that window and pass a truncated buffer to sharp, producing a "corrupt header" error that points at the image asset rather than the cache.

Fix: write to a uniquely-named temp file (.id.XXXXXX) in the same directory, then rename() to the final path. rename() is a single atomic syscall. Concurrent readers will only see a complete file.

The temp file follows the same convention as rsync: a dot-prefixed name with a random suffix, written in the same destination dir to ensure that the filesystem stays the same.

See: https://man7.org/linux/man-pages/man1/rsync.1.html

  PR?)

It does not

- **Other information**:

We discovered this issue because we run several React builds in parallel with vite-imagetools. Builds turned flaky on CI after we moved to vite-imagetools, but never failed when run alone. The error surfaces as a build-time plugin error pointing at the image asset:

```
Error: [imagetools] Could not load src/assets/backgrounds/foo.webp?imagetools (imported by src/themes/bar.ts): Input file contains unsupported image format
```

I didn't use os.tmpdir() because on Linux /tmp is commonly a separate tmpfs mount, and rename() across filesystems fails with EXDEV.

I also have a test case which I haven't included since it needs to mock node's `writeFile`. This makes the test a bit complicated to understand but I am happy to include it if you want to.

### DIFF
--- a/.changeset/fix-cache-atomic-write.md
+++ b/.changeset/fix-cache-atomic-write.md
@@ -1,0 +1,18 @@
+---
+"vite-imagetools": patch
+---
+
+fix: use atomic write for cache files to prevent corrupt reads
+
+Node's writeFile opens the target with O_TRUNC, zeroing it before the
+data arrives. A concurrent build process can read that window and pass
+a truncated buffer to sharp, producing a "corrupt header" error that
+points at the image asset rather than the cache.
+
+Fix: write to a uniquely-named temp file (.id.XXXXXX) in the same
+directory, then rename() to the final path. rename() is a single
+atomic syscall. Concurrent readers will only see a complete file.
+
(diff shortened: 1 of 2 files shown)


nickgarlis:
Quick update: we’ve patched the dependency to include this fix and have had no build issues since. @benmccann can you PTAL?

benmccann:
Thanks. Closing in favor of https://github.com/JonasKruckenberg/imagetools/pull/915 which also added some tests

PR #902, opened 2026-06-17
chore: add vite package bugs metadata

BowenMilner:
## Summary
- add bugs.url metadata to the vite-imagetools package manifest
- leave runtime code, dependencies, entry points, and package files unchanged

## Validation
- node package metadata assertion
- npm pack --dry-run --ignore-scripts --json --cache /private/tmp/codex-npm-cache-imagetools-vite
- git diff --check

### DIFF
--- a/.changeset/add-vite-bugs-metadata.md
+++ b/.changeset/add-vite-bugs-metadata.md
@@ -1,0 +1,5 @@
+---
+"vite-imagetools": patch
+---
+
+Added package metadata for the issue tracker.
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -18,6 +18,9 @@
     "type": "git",
     "url": "https://github.com/JonasKruckenberg/imagetools.git"
   },
+  "bugs": {
+    "url": "https://github.com/JonasKruckenberg/imagetools/issues"
+  },
   "keywords": [
     "resize-images",
     "responsive-images",


benmccann:
What's the motivation for this change?

BowenMilner:
The motivation is to make the published `vite-imagetools` package expose its issue tracker through standard npm package metadata. npm, package managers, scanners, and registry UIs can read the `bugs.url` field directly, which makes it easier for users and tooling to route bug reports to the existing GitHub issues page without changing runtime behavior.\n\nI added a patch changeset because this changes the published package metadata only; no source code or API behavior is affected.

benmccann:
What tool specifically are you trying to use that's looking for the field? I've never run into one that uses it. The repo URL is already included which makes it pretty obvious what the issue URL is

BowenMilner:
Thanks for taking a look. The concrete consumer I had in mind is npm/package metadata tooling rather than a human reading the package page. `bugs.url` is a standard package.json field that npm-oriented tools can read directly when they want the package's issue/support URL, while `repository.url` only says where the source lives.

For example, npm metadata consumers, package inventory/audit scripts, and registry UIs can surface `bugs.url` as the issue tracker without needing to infer `https://github.com/.../issues` from the repository URL. The repo URL is definitely enough for someone browsing manually; this PR is just making the issue tracker explicit in the package metadata so `vite-imagetools` matches the common npm package shape.

That said, if you prefer not to carry metadata-only changes like this, no worries from my side.

benmccann:
I get that it would be used in an automated fashion, but I'm wondering by what specifically?

I'm not against this change in general, but I am a bit worried that it's just churn/spam potentially driven by some weird crypto AI thing and don't want to rewards changes that are made for spurious reasons

BowenMilner:
Thanks, that concern makes sense. To be concrete, the examples I was thinking of are npm metadata consumers rather than the GitHub UI:

- `npm bugs vite-imagetools` reads the package `bugs` metadata and opens/reports the configured issue tracker.
- `npm view vite-imagetools bugs --json` exposes the field directly for package inventory/audit scripts.
- Internal dependency dashboards and registry/package indexers often treat `repository` and `bugs` as separate fields: source location vs. issue/support location.

So I agree the repository URL is enough for a human to infer the issue tracker. The value of this change is just making the issue URL explicit for tooling that reads package metadata literally instead of guessing that the support URL is `${repository}/issues`.

On the bounty/AI concern: this is not intended to pressure maintainers into accepting churn. It is a very small metadata-only cleanup, and if you do not think it is useful enough for this package, I am completely fine with closing it. I opened it because it aligns the published package manifest with a standard npm field, not because there is any runtime bug behind it.

JonasKruckenberg:
lets just close this. the npm registry info is plenty for human (and even just somewhat smart scanners). lets not make it easier for low effort scrapers to hassle us

PR #904, opened 2026-06-19
fix: recompute pixelDensityDescriptor on cache hit

### LINKED ISSUE #751: Directive `basePixels` does not work as expected with React and SSR
First of all, thanks for this great package. I was able to generate a fully reponsive picture with only a few lines of code, which is amazing! To the problem

I am trying to use the `basePixels` directive with React, but it does not work as expected. When rendered on the server, I get the correct 1x and 2x images, but when it renders on the client, it switches to 300w and 600w versions, and I get a hydration mismatch. 

Packages
- vite-imagetools 7.0.4
- React canary 19.0.0-rc-1eaccd82-20240816
- Remix 2.12.1
- Vite 5.1.0

This is my component:

```tsx
import square2x from '~/image/services.jpg?w=300;600&basePixels=300&format=avif;webp;jpeg&as=picture'

console.log(square2x)

const formats = ['avif', 'webp', 'jpeg']

export default function ViteImageTools() {
  return (
    <main>
      <h1>vite-imagetool</h1>
      <div className="my-6 w-[300px]">
        <picture>
          {formats.map((f) => (
            <source key={f} srcSet={square2x.sources[f]} type={`image/${f}`} />
          ))}
          <img
            src={square2x.img.src}
            alt="Testing vite-imagetools"
            width={square2x.img.w}
            height={square2x.img.h}
          />
        </picture>
      </div>
    </main>
  )
}
```

This is the output in the linux console (SSR)

```js
{
  sources: {
    avif: '/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc 1x, /@imagetools/16bf4d6638f522afb34cd2b0322ce3629a686963 2x',
    webp: '/@imagetools/a9ce3b8e03e646d59be2173e1e8a5ac7b4757714 1x, /@imagetools/8c6aafbb3319208df40ff436fabfc501426a5684 2x',
    jpeg: '/@imagetools/b14e2bce7ff2b46c1553fd7dfeb18b7d3659eeae 1x, /@imagetools/f9ebb33f5f5802dcf07f93c05a54f3068167bd29 2x'
  },
  img: {
    src: '/@imagetools/f9ebb33f5f5802dcf07f93c05a54f3068167bd29',
    w: 600,
    h: 600
  }
}
```

And the output in the browser console

```js
{
  "sources": {
    "avif": "/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc 300w, /@imagetools/16bf4d6638f522afb34cd2b0322ce3629a686963 600w",
    "webp": "/@imagetools/a9ce3b8e03e646d59be2173e1e8a5ac7b4757714 300w, /@imagetools/8c6aafbb3319208df40ff436fabfc501426a5684 600w",
    "jpeg": "/@imagetools/b14e2bce7ff2b46c1553fd7dfeb18b7d3659eeae 300w, /@imagetools/f9ebb33f5f5802dcf07f93c05a54f3068167bd29 600w"
  },
  "img": {
    "src": "/@imagetools/f9ebb33f5f5802dcf07f93c05a54f3068167bd29",
    "w": 600,
    "h": 600
  }
}
```

React complains with

```js
A tree hydrated but some attributes of the server rendered HTML didn't match the client properties. This won't be patched up. This can happen if a SSR-ed Client Component used:

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

https://react.dev/link/hydration-mismatch

 <RenderedRoute match={{params:{}, ...}} routeContext={{...}}>
      <Layout>
        <html lang="en">
          <head>
          <body>
            <App>
              <Outlet>
                <RenderedRoute match={{params:{}, ...}} routeContext={{outlet:null, ...}}>
                  <ViteImageTool>
                    <main>
                      <h1>
                      <div className="my-6 w-[30...">
                        <picture>
                          <source
+                           srcSet="/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc 300w, /@imagetools/16bf4d663..."
-                           srcSet="/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc 1x, /@imagetools/16bf4d6638f..."
                            type="image/avif"
                          >
                          <source
+                           srcSet="/@imagetools/a9ce3b8e03e646d59be2173e1e8a5ac7b4757714 300w, /@imagetools/8c6aafbb3..."
-                           srcSet="/@imagetools/a9ce3b8e03e646d59be2173e1e8a5ac7b4757714 1x, /@imagetools/8c6aafbb331..."
                            type="image/webp"
                          >
                          <source
+                           srcSet="/@imagetools/b14e2bce7ff2b46c1553fd7dfeb18b7d3659eeae 300w, /@imagetools/f9ebb33f5..."
-                           srcSet="/@imagetools/b14e2bce7ff2b46c1553fd7dfeb18b7d3659eeae 1x, /@imagetools/f9ebb33f5f5..."
                            type="image/jpeg"
                          >
```

To be able to debug this problem, I forked and installed a local copy of `vite-imagetools`. I put a `console.log(metadatas)` into `output-format.tsx` > [pictureFormat](https://github.com/JonasKruckenberg/imagetools/blob/074f9f7a2a69e8a480270f44e820587026773bbc/packages/core/src/output-formats.ts#L76) function and I see, that second log is missing the `pixelDensityDescriptor`.

First metadata contains `pixelDensityDescriptor`:

```js
[
  {
    format: 'avif',
    width: 300,
    height: 300,
    space: 'srgb',
    channels: 3,
    depth: 'uchar',
    density: 300,
    chromaSubsampling: '4:4:4',
    isProgressive: true,
    resolutionUnit: 'inch',
    hasProfile: true,
    hasAlpha: false,
    orientation: 1,
    aspect: 1,
    allowUpscale: false,
    pixelDensityDescriptor: '1x',
    src: '/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc',
    image: undefined
  },
  //...
]
```

Second data does not:

```js
[
  {
    format: 'avif',
    width: 300,
    height: 300,
    space: 'srgb',
    channels: 3,
    depth: 'uchar',
    isProgressive: false,
    pages: 1,
    pagePrimary: 0,
    compression: 'av1',
    hasProfile: false,
    hasAlpha: false,
    src: '/@imagetools/6b96b93013bb7a72f4ff6f967dbd76974b92bfbc',
    image: undefined
  },
  //...
]
```

Not sure if this is a bug or I am doing something wrong ... I think it might be a caching problem or maybe some weird React/Remix race condition. I can create the reproduction repo, when I get back from vacation in a few days ;)

blt-r:
Closes #751.

When reading from cache, applyTransforms is never called so pixelDensityDescriptor was always undefined. This caused srcset to fall back to width descriptors instead of density descriptors.

Compute it inline from imageConfig.basePixels and metadata.width using the same formula as the resize transform, for backwards compatibility with the existing metadata interface.

This is a bit hacky, but I think this is the only way to fix the bug without a braking change.

### DIFF
--- a/.changeset/fix-pixel-density-cache.md
+++ b/.changeset/fix-pixel-density-cache.md
@@ -1,0 +1,5 @@
+---
+"vite-imagetools": patch
+---
+
+Fix pixelDensityDescriptor not being set when reading from cache.
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -159,6 +159,11 @@ export function imagetools(userOptions: Partial<VitePluginOptions> = {}): Plugin
             // value so emitted filenames don't change between cache misses and cache hits.
             if (typeof imageConfig.format === 'string' && metadata.format !== imageConfig.format)
               metadata.format = imageConfig.format as ImageMetadata['format']
+            // recompute pixelDensityDescriptor from config and the actual resized width.
+            // ImageConfig is typed as string | string[] but resolveConfigs always produces
+            // single string values via cartesian product, so this is always a string.
+            const parsedBasePixels = parseInt(String(imageConfig.basePixels || ''))
+            metadata.pixelDensityDescriptor = parsedBasePixels > 0 ? `${metadata.width / parsedBasePixels}x` : undefined
           } else {
             const { transforms } = generateTransforms(imageConfig, transformFactories, srcURL.searchParams, logger)
             const res = await applyTransforms(transforms, img.clone(), pluginOptions.removeMetadata)


benmccann:
This PR will need a rebase

> This is a bit hacky, but I think this is the only way to fix the bug without a braking change.

Can you describe what the cleaner fix would be if you were allowed a breaking change?

blt-r:
> Can you describe what the cleaner fix would be if you were allowed a breaking change?

I think the actual images metadata that is in the image itself (sharp's `Metadata`) should be decoupled from the data about transforms and output formats (the additional fields on `ImageMetadata`). Currently, on a cache hit none of these fields are populated at all, which happens to only matter for `pixelDensityDescriptor`, because it's the only one used not for creating the image itself but other things in the output.

Alternatively, a simpler fix that's cleaner then this PR would be to just remove the `pixelDensityDescriptor` from `ImageMetadata` and compute it on the spot when needed. And then we should only have fields in `ImageMetadata` that don't don't matter as long as we already have the image. That would still be a breaking change since it's a public field.

benmccann:
I'm not fully sure I understand the optimal solution, but would you be able to throw up a PR for it? I just made a breaking change so we could batch the two together in a single release

PR #909, opened 2026-07-15
fix: keep format directive value on cache hits so emitted filenames are stable

### LINKED ISSUE #908: `?format=jpg` emits `.jpg` on cache miss but `.jpeg` on cache hit — unstable filenames and dangling asset references
**Package**: vite-imagetools 10.0.1 (imagetools-core 10.0.0)

## Describe the bug

With the build cache enabled (the default), an import like

```ts
import hero from './hero.jpg?w=480;768&format=avif;webp;jpg&as=picture'
```

emits the JPEG variants with a different file extension depending on cache state:

- **Cache miss**: the emitted filename extension comes from `metadata.format`, which the `format` transform sets to the raw directive value → `hero-<hash>.jpg`
- **Cache hit**: metadata is re-derived via `sharp(cachedBuffer).metadata()`, and sharp reports the format as `jpeg` → `hero-<hash>.jpeg`

The bytes (and therefore Vite's content hashes) are identical in both cases — only the extension flips.

This is the same class of problem that `executeTransform` already works around for avif (sharp reads cached avif back as `heif`/`av1`, see the comment referencing sharp#2504/#3746) — `jpg` vs `jpeg` is just the alias pair that isn't handled.

## Why this matters

Any two builds that share a cache directory can disagree on asset filenames. The nastiest case is an SSR framework (TanStack Start, SvelteKit, etc.) running a client build followed by a server build in the same working directory:

1. Client build: cache is cold → misses → emits `hero-<hash>.jpg`, writes the cache
2. SSR build: cache is now warm → hits → resolves the same imports to `hero-<hash>.jpeg`

The server bundle now references `.jpeg` asset URLs that the client build never emitted — every JPEG `srcset` fallback 404s for consumers that don't pick the avif/webp sources (legacy browsers, crawlers harvesting `srcset`). We hit this in production; the emitted variant content hashes matched across builds exactly as described, with only the extension differing.

It also breaks build reproducibility in general: warm-cache and cold-cache builds of the same source produce differently-named artifacts.

## Reproduction

Two consecutive `vite build` runs sharing the default cache dir, with any `?format=jpg` import:

```ts
import img from './photo.png?format=jpg'
```

- Build 1 (cold cache): dist contains `photo-<hash>.jpg`
- Build 2 (warm cache): dist contains `photo-<hash>.jpeg`

## Expected behavior

The emitted filename is independent of cache state: `?format=jpg` always emits `.jpg`.

## Suggested fix

In `executeTransform`'s cache-hit branch, restore the directive's format whenever sharp's re-derived value differs — generalizing the existing avif workaround:

```ts
if (typeof imageConfig.format === 'string' && metadata.format !== imageConfig.format)
  metadata.format = imageConfig.format as ImageMetadata['format']
```

This makes the cache-hit path produce exactly the metadata the cache-miss path stores (the miss path always keeps the directive value), so it can't change behavior for any format where the two already agree. PR with a regression test incoming.

niantang-2026:
- **Quick Checklist**

Bug fix. Fixes #908.

With the build cache enabled (the default), `?format=jpg` imports emit `.jpg` variants on a cache miss but `.jpeg` variants on a cache hit: the miss path stores the raw directive value in `metadata.format`, while the hit path re-derives metadata via `sharp(cachedBuffer).metadata()`, and sharp reports JPEG as `jpeg`. Since the emitted asset filename extension comes from `metadata.format`, builds sharing a cache directory disagree on asset names. In an SSR setup (client build then server build in the same working directory) the server bundle ends up referencing `.jpeg` URLs the client build never emitted — dangling references that 404 at runtime.

This is the same cache-restore metadata mismatch already special-cased for avif (sharp reads cached avif back as `heif`/`av1`). This PR generalizes that workaround: on a cache hit, restore the directive's format whenever sharp's re-derived value differs. The cache-miss path always keeps the directive value, so this makes the hit path produce exactly the metadata the miss path would — it cannot change behavior for formats where the two already agree, and the existing avif test still passes.

Not a feature change. Emitted filenames are now independent of cache state: `?format=jpg` always emits `.jpg`, on cold and warm caches alike.

No. Projects that (inadvertently) relied on warm-cache builds emitting `.jpeg` for `?format=jpg` will see the extension settle on `.jpg`, which is what a clean build always produced.

- **Other information**:

Note that `metadata.format` also feeds data-URI/`Content-Type` strings (`image/jpg` for `?format=jpg&inline`); that was already the cache-miss behavior before this PR, so this change only removes the hit/miss divergence rather than introducing a new format name anywhere.

The regression test builds twice against a shared cache dir and asserts the `.jpg` asset exists with an identical filename in both bundles; before the fix the warm-cache build emits `.jpeg` and the test fails.

### DIFF
--- a/.changeset/spotty-mails-jump.md
+++ b/.changeset/spotty-mails-jump.md
@@ -2,4 +2,4 @@
 'vite-imagetools': patch
 ---
 
-Keep the `format` directive's value when restoring images from the build cache, so emitted asset filenames no longer flip between `.jpg` (cache miss) and `.jpeg` (cache hit). Generalizes the existing avif/heif cache workaround to all formats whose sharp-reported name differs from the directive value.
+fix: keep the `format` directive's value when restoring images from the build cache, so emitted asset filenames no longer flip between `.jpg` and `.jpeg`
--- a/.changeset/spotty-mails-jump.md
+++ b/.changeset/spotty-mails-jump.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+Keep the `format` directive's value when restoring images from the build cache, so emitted asset filenames no longer flip between `.jpg` (cache miss) and `.jpeg` (cache hit). Generalizes the existing avif/heif cache workaround to all formats whose sharp-reported name differs from the directive value.
--- a/packages/vite/src/__tests__/main.test.ts
+++ b/packages/vite/src/__tests__/main.test.ts
@@ -565,6 +565,33 @@ describe('vite-imagetools', () => {
         expect(files).toHaveLength(1)
       })
     })
+
+    describe('cache.jpgFormat', () => {
+      test('keeps the jpg extension on cache hits', async () => {
+        const dir = './node_modules/.cache/imagetools_test_cache_jpg_format'
+        await rm(dir, { recursive: true, force: true })
+        const buildOnce = async () =>
+          (await build({
+            root: join(__dirname, '__fixtures__'),
+            logLevel: 'warn',
+            build: { write: false },
+            plugins: [
+              testEntry(`
+                            import Image from "./pexels-allec-gomes-5195763.png?format=jpg"
+                            window.__IMAGE__ = Image
+                        `),
+              imagetools({ cache: { dir } })
+            ]
+          })) as RollupOutput | RollupOutput[]
+
+        const coldFiles = getFiles(await buildOnce(), '**.jpg') as OutputAsset[]
+        const warmFiles = getFiles(await buildOnce(), '**.jpg') as OutputAsset[]
+
+        expect(coldFiles).toHaveLength(1)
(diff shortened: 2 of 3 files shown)


.changeset/spotty-mails-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+Keep the `format` directive's value when restoring images from the build cache, so emitted asset filenames no longer flip between `.jpg` (cache miss) and `.jpeg` (cache hit). Generalizes the existing avif/heif cache workaround to all formats whose sharp-reported name differs from the directive value.
```
benmccann:
```suggestion
fix: keep the `format` directive's value when restoring images from the build cache, so emitted asset filenames no longer flip between `.jpg` and `.jpeg`
```

benmccann (approved):

PR #915, opened 2026-07-31
fix(vite): write build cache entries atomically

v-anton:
The build cache is written with a plain `writeFile()`:

```js
cachedBuffer = await image.toBuffer()
await writeFile(`${cacheOptions.dir}/${id}`, cachedBuffer)
```

`writeFile()` truncates the target and then fills it, so the entry is observable in a partial state for the duration of the write. Any write that doesn't run to completion leaves a truncated file at the cache path:

- a build interrupted partway (Ctrl-C, CI timeout, OOM kill)
- two builds sharing `node_modules/.cache/imagetools` — `synchronizedTransform` dedupes by id, but that map is per-process, so it doesn't serialise anything across processes (parallel CI jobs, a monorepo building several packages, a build racing a test run)
- the write itself failing under load — asset-heavy projects write thousands of entries concurrently, and an `EMFILE`/`ENOSPC`/transient I/O error partway through leaves the bytes already flushed

The read path only guards on size:

```js
(statSync(`${cacheOptions.dir}/${id}`, { throwIfNoEntry: false })?.size ?? 0) > 0
```

A truncated file passes that guard, so a later build treats it as a valid entry. Two outcomes, and the quiet one is the worse one:

- **Truncated after the header** — `sharp(cachedBuffer).metadata()` succeeds, because metadata only reads the header. The truncated buffer is then handed to `emitFile()`. **A corrupt image is written to the output and the build exits 0.** I hit this on a real project: a 314 KB source PNG shipped as a 64-byte file, no error anywhere.
- **Truncated before the header** — the build fails with `Input buffer contains unsupported image format`.

Both are sticky: the poisoned entry survives until someone manually clears the cache directory, which is why this tends to show up as an intermittent, hard-to-reproduce crash on asset-heavy builds rather than something tied to an obvious trigger.

### Fix

Stage the entry under a unique temp name and `rename()` it into place. `rename(2)` is atomic within a filesystem, so a cache entry is either absent or complete — a partial write is never observable at the target path. The temp name includes the pid, so concurrent builds sharing a cache directory stage to distinct files and the last `rename` simply wins. The temp file is removed if the write fails.

This only changes how the bytes land; cache keys, layout, and the read path are untouched.

### Tests

Two tests in the existing `utils` block:

- entries are written in full and no `.tmp` files are left behind
- a concurrent reader never observes a partially written file

The second is the regression test. Against the previous `writeFile()` implementation it fails with:

```
AssertionError: expected [ +0, 524288, 1572864, 4194304 ] to deeply equal [ 4194304 ]
```

— a reader catching the target at 0 bytes, 512 KB and 1.5 MB mid-write. Those are exactly the states that pass the `size > 0` guard.

### Note on the suite

`packages/vite` has 27 failing tests on `main` at efe000b, unrelated to this change (mostly the `logging` and rollup-error assertions). This PR takes it from 13 passed / 27 failed to 15 passed / 27 failed.

### Follow-up worth considering

The read path could also be hardened — wrapping the cache-hit decode and falling through to a re-transform on failure would let already-corrupted caches self-heal instead of needing a manual `rm -rf`. I left it out to keep this focused on the root cause; happy to add it here if you'd prefer.

### DIFF
--- a/.changeset/nervous-cameras-repeat.md
+++ b/.changeset/nervous-cameras-repeat.md
@@ -1,0 +1,5 @@
+---
+'vite-imagetools': patch
+---
+
+fix: write build cache entries atomically, so an interrupted build can't leave a truncated entry that later builds read back as valid
--- a/packages/vite/src/__tests__/main.test.ts
+++ b/packages/vite/src/__tests__/main.test.ts
@@ -7,9 +7,9 @@ import { type OutputAsset, type OutputChunk, type RollupOutput } from 'rollup'
 import { JSDOM } from 'jsdom'
 import sharp from 'sharp'
 import { afterEach, describe, test, expect, it, vi } from 'vitest'
-import { createBasePath } from '../utils'
+import { createBasePath, writeFileAtomic } from '../utils'
 import { existsSync } from 'node:fs'
-import { rm, utimes, readdir, copyFile } from 'node:fs/promises'
+import { rm, utimes, readdir, copyFile, mkdir, readFile } from 'node:fs/promises'
 
 expect.extend({ toMatchImageSnapshot })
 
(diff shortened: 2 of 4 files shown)


benmccann:
Any difference between this PR and https://github.com/JonasKruckenberg/imagetools/pull/886 ?

I'm on mobile so haven't compared them

benmccann:
> packages/vite has 27 failing tests on main at https://github.com/JonasKruckenberg/imagetools/commit/efe000b44a9ea912e64566e99f8feccf4d69586e, unrelated to this change (mostly the logging and rollup-error assertions).

Nothing is failing on `main`: https://github.com/JonasKruckenberg/imagetools/commits/main/. It must be something specific to your machine. Are you on Windows perhaps?

PR #916, opened 2026-08-01
refactor: compose `ImageMetadata` into `info` and `transforms`

blt-r:
In relation to https://github.com/JonasKruckenberg/imagetools/pull/904#issuecomment-5136640333.

### Summary

This PR replaces the flat, `sharp`-driven `ImageMetadata` with a composed, typed shape and threads it through the transform pipeline as explicit state. It removes the `METADATA` symbol plumbing, drops the stale `pixelDensityDescriptor` field, renames three types to match their roles, and keeps the `as=metadata` output byte-for-byte backward compatible.

---

### The old code

`ImageMetadata` was a flat object that **extended sharp's `Metadata`** and carried an **index signature**:

```ts
interface ImageMetadata extends Metadata {
  [key: string]: unknown   // anything, unchecked
  src: string
  width: number
  // ...dozens of sharp fields mixed with pipeline state
  pixelDensityDescriptor?: string
}
```

State was persisted **onto the sharp instance itself** through a hidden symbol:

```ts
export const METADATA = Symbol('image metadata')
declare module 'sharp' {
  interface Sharp { [METADATA]: ImageMetadata }
}
export function setMetadata(image: Sharp, key: string, value: unknown) { ... }
```

Every transform did `image[METADATA].quality = 80` and every helper did `getMetadata(image, 'format')`. The pipeline's data lived on the wrong object.

### Why it was bad

- **No type safety.** The index signature let any string key through unchecked. Custom transforms could silently set or read misspelled keys with zero feedback from the compiler.
- **State coupled to the sharp instance lifecycle.** The recorded values were only reachable while the sharp object lived, and only through a symbol that custom transforms had to reach into. There was no way to persist, inspect, or reconstruct what happened to an image after the fact.
- **Cache hits lost everything.** When the vite plugin restored an image from its build cache, the sharp instance was rebuilt from the encoded file, so `image[METADATA]` was empty — every applied-transform value was gone. The plugin papered over this by reading back `image.metadata()` and reconciling a couple of fields, but the rest (flip, quality, rotate, ...) silently vanished.
- **Dimensions lied.** `width`/`height` recorded the *targets declared by the transforms*, never the actual rendered output. `rotate` swaps width and height but never updated them, so metadata could disagree with the image actually being served.
- **`pixelDensityDescriptor` was a cached derivative.** It was computed once at transform time from the declared width and then frozen onto the metadata. On a cache hit it was stale, and whenever `rotate` changed the rendered dimensions it produced wrong descriptors (`300w` instead of `2x`).
- **Ownership was muddled.** Pipeline state, sharp's read-only metadata, the source URL, the config, and the image object all lived on one flat bag, duplicated across layers with no clear owner.

---

### The new code

`ImageMetadata` is now a small, composed, fully typed object:

```ts
interface ImageMetadata {
  info: { width: number; height: number; autoOrient: { width: number; height: number } }
  transforms: AppliedTransforms   // typed optional fields, only applied transforms are present
}
```

Transforms record what they did into `state.transforms` and return the transformed sharp instance:

```ts
return function flipTransform(state, image) {
  state.transforms.flip = true
  return image.flip()
}
```

`applyTransforms` builds the initial state from the source file and threads it through the pipeline (as the first argument to each `ImageTransformation`, and the second to each `TransformOption`), strips metadata when `removeMetadata` is enabled, and returns an `ApplyTransformsResult`:

```ts
{ image, metadata, raw }
```

The vite plugin consumes a `ProcessedImage` (`{ src, image, config, info, transforms, raw }`) and reconciles `info.width/height` against the encoded output on *every* transformation, not just cache misses.

### Migrating custom transforms

Custom transforms previously stored shared state on the sharp instance; they now write to `state.transforms` instead:

```ts
// before
return (image) => {
  image[METADATA].myCustomValue = 'x'
  return image
}

// after
return (state, image) => {
  state.transforms.myCustomValue = 'x'
  return image
}
```

For typed access, extend `AppliedTransforms` via declaration merging (see below).

### Why it's good and needed

- **Type safety.** No index signature, no `unknown` leakage. Misspelled keys and wrong value types are compile errors now.
- **Extensibility preserved.** The old index signature was also how custom transforms stashed arbitrary shared state (`image[METADATA].foo = x`). That role now falls to `AppliedTransforms`, which is an exported `interface` that consumers can extend via declaration merging:

  ```ts
  declare module 'imagetools-core' {
    interface AppliedTransforms {
      myCustomValue?: string
    }
  }
  ```

  Custom transforms record on `state.transforms.myCustomValue` and custom output formats read it back — fully typed, and included in the `as=metadata` output without reintroducing `unknown` anywhere.
- **State is data, not a side effect.** Applied-transform values can be inspected, serialized, and rebuilt — the foundation for correctly reconstructing metadata on cache hits instead of losing it.
- **Honest dimensions.** Metadata reflects the image that is actually served, so srcsets and density descriptors always match. `basePixels` now yields `2x` even with a `rotate` directive or without any `w`/`h`/`aspect` directive.
- **Clean ownership.** `raw` (sharp's read-only metadata), `transforms` (what we did), `info` (dimensions as they flow), `config` (how it was generated) are separated — no circularity, no duplication.
- **Backward compatible where it matters.** The `as=metadata` output keeps its exact previous flat shape; verified byte-for-byte against the old build across directive combinations, including the cache-hit path. Breaking changes are confined to the plugin/core API surface (custom transforms, output formats, types) and are documented in the changeset.

### Testing

- 290 core + 44 vite tests green; `pnpm lint` clean; `tsc --noEmit` passes.
- Old vs. new `as=metadata` output byte-diffed across ~9 directive combinations (cold and warm cache); the non-whitelist output is identical.

---

### Vibe code alert

I haven't read any of the new or old test, but the ai seems sure that they are good. But the rest of the code should be definitely good.

### Quick Checklist


benmccann:
Can you drop the docs changes? There are some complications and I periodically regenerate them after a batch of PRs

benmccann:
thanks!

merged in https://github.com/JonasKruckenberg/imagetools/pull/918. please leave the "allow edits from maintainers" box changed as I could have merged this PR then

PR #921, opened 2026-08-04
Fix a batch of caching problems

blt-r:
I wanted to make this PR before 11.0.0 was out and slip all these breaking changes in there, but im late so whatever

## Summary

The image cache is the plugin's source of truth for both the bytes it serves and the metadata a `metadata` import reports, but it was never keyed or populated consistently. This PR is a sweep of cache fixes: the cache id mixed in output-format params that cannot change the bytes, `as=metadata` imports spun up their own redundant cache entries, and the metadata they returned differed depending on whether the image was served from the cache or processed fresh.

## 1. Cache ids are keyed only by image directives

- `ImageConfig` is now `Record<string, string>` (single-string directives only, never arrays); `resolveConfigs` still returns the same `ImageConfig[]` shape, so existing custom `resolveConfigs` overrides keep working unchanged. Because the config now holds only directives that affect the bytes, the cache id derived from it finally matches the data.
- Transform options now see `string | undefined` instead of `string | string[] | undefined`. The `string[]` branches are gone from `background`, `tint`, `parseBasePixels`, and the output-format restore path; the runtime guards stay, so a custom resolver returning a non-string is skipped rather than mis-applied.

## 2. Serialization params no longer split the cache

- Output-format params and the `as` selector are excluded from the configs and from the cartesian product. Since the cache id hashes only the directives, URLs that differ only in `as`/`metadata=` collapse into one cache entry instead of duplicating identical sharp work and disk. `generateImageID` itself is unchanged; the change comes entirely from the configs no longer containing those params.

## 3. `as=metadata` is deterministic across cache hits and misses

- **Applied-transform directives are no longer exported.** They were just the inputs, an echo of what the user typed into the import URL, so exporting them back wasn't very meaningful anyways. And they just didn't work with cache at all. It may be possible to either cache them or reconstruct them but I don't see value in that.
- **The cache-miss path reads the encoded output's metadata.** instead of the source image's, so a miss reports exactly what the hit path reads back from the file: `isProgressive`, `hasAlpha`, `pages`, `density`, etc. never differ.

## Smaller details

- A custom output format whose name collides with a directive name (e.g. a format named `blur`) suppresses that directive, since every key in `outputFormats` is excluded from the configs. This is documented in the `resolveConfigs` JSDoc and the changeset; name custom formats to avoid directive names.

## Breaking changes

- `ImageConfig` values are single strings, never arrays.
- Transform options are `string | undefined`, not `string | string[] | undefined`.
- `as=metadata` imports no longer include the applied-transform directive fields.
- Cache ids change for imports carrying `as`/output-format params (they collapse to the directive-only id).

## Testing

- core (295): `resolveConfigs` product/exclusion cases, `as` exclusion, custom output-format name suppressing a directive, string-only values, metadata output without transform directives, `:` guard.
- vite (50): shared cache entry across `metadata=width`, `metadata=width;height`, `as=metadata`, `as=url`, `as=srcset`, `as=picture`; custom `resolveConfigs` override; metadata omitting transform directives on both a cold and a warm cache; metadata identical across a cache hit and miss.
- `pnpm lint` and `tsc --noEmit` are clean.

## Quick Checklist

### DIFF
--- a/packages/core/src/__tests__/output-formats.test.ts
+++ b/packages/core/src/__tests__/output-formats.test.ts
@@ -88,11 +88,6 @@ describe('metadata format', () => {
       height: 150,
       format: 'webp'
     })
-    expect(output).not.toHaveProperty('flip')
-    expect(output).not.toHaveProperty('quality')
-    expect(output).not.toHaveProperty('rotate')
-    expect(output).not.toHaveProperty('tint')
-    expect(output).not.toHaveProperty('fit')
   })
 
   test('whitelist', () => {
--- a/packages/core/src/__tests__/output-formats.test.ts
+++ b/packages/core/src/__tests__/output-formats.test.ts
@@ -7,7 +7,7 @@ function meta(
   width: number,
   height = width,
   format?: string,
-  config: Record<string, string | string[]> = {},
+  config: Record<string, string> = {},
   raw: Partial<Metadata> = {},
   transforms: Partial<AppliedTransforms> = {}
 ): ProcessedImage {
--- a/packages/core/src/__tests__/output-formats.test.ts
+++ b/packages/core/src/__tests__/output-formats.test.ts
@@ -77,7 +77,7 @@ describe('metadata format', () => {
     ])
   })
 
-  test('includes the values threaded through the transforms', () => {
+  test('does not include the values threaded through the transforms', () => {
     const output = metadataFormat()([
       meta('/foo.jpg', 300, 150, 'webp', {}, {}, { flip: true, quality: 80, rotate: 90, tint: '#ff0000', fit: 'cover' })
     ])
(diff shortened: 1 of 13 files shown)


packages/core/src/__tests__/output-formats.test.ts
```diff
@@ -86,13 +86,13 @@ describe('metadata format', () => {
...
-      rotate: 90,
-      tint: '#ff0000',
-      fit: 'cover'
+      format: 'webp'
     })
+    expect(output).not.toHaveProperty('flip')
+    expect(output).not.toHaveProperty('quality')
+    expect(output).not.toHaveProperty('rotate')
+    expect(output).not.toHaveProperty('tint')
+    expect(output).not.toHaveProperty('fit')
```
benmccann: I don't think these are necessary. If they were present I believe the above `toMatch` assertion would fail
```suggestion
```

benmccann (approved):

